PR #1, opened 2019-07-23
Support diego cluster creation

thardeck:


scripts/include/common.sh
```diff
@@ -19,9 +19,8 @@ export DOCKER_ORG="${DOCKER_ORG:-cap}"
...
 export DOCKER_PASSWORD="${DOCKER_PASSWORD:-}"
 export ENABLE_EIRINI="${ENABLE_EIRINI:-true}"
-export CREATE_EIRINI_NAMESPACE="${CREATE_EIRINI_NAMESPACE:-false}"
 
 
 if [ -z "${DEFAULT_STACK}" ]; then
-  export DEFAULT_STACK=$(helm inspect helm/cf/ | grep DEFAULT_STACK | sed  's~DEFAULT_STACK:~~g' | sed 's~"~~g' | sed 's~\s~~g')
-fi
\ No newline at end of file
+  export DEFAULT_STACK=$(helm inspect suse/cf | grep DEFAULT_STACK | sed  's~DEFAULT_STACK:~~g' | sed 's~"~~g' | sed 's~\s~~g')
```
mudler: Not sure about this, in this way we get the default stack only if the release was deployed in the cluster, right? ( before we were inspecting the local chart, so it could have been resolved without a deployed release)

scripts/setup.sh
```diff
@@ -122,11 +122,13 @@ kubectl delete storageclass standard
 kubectl create -f ../kube/storageclass.yaml
 helm init --upgrade --wait
 
-if [ "$CREATE_EIRINI_NAMESPACE" = true ] ; then
+if [ "$ENABLE_EIRINI" = true ] ; then
   kubectl create namespace eirini
```
mudler: They can't be tied together, it is separate now just because between the eirini-bosh release changes we do create the namespace now ( and in the future we can drop the creation at all). 

In the new eirini bosh release(>0.0.4)  certs are copied in the namespace if the namespace didn't existed before ( so to be idempotent, we should copy the certs required for the uploader)

mudler: Oh wait true, nevermind my prior comment, this is ok. If we create it always should be ok as well

PR #2, opened 2019-07-31
Let the user specify the cluster name and respect that in all make ta…

jimmykarily:
…rgets

Signed-off-by: Dimitris Karakasilis <DKarakasilis@suse.com>


jimmykarily:
This allow us to create multiple clusters are the same time without "Make" messing around with each other's container.

mudler:
That's really nice! thank you!

PR #4, opened 2019-08-30
Add functionality to deploy caasp4 on openstack

viccuad:
This PR adds scripts and content to deploy caasp4 on openstack.
See `make all-caasp4os`.

It creates a docker image with the caasp4 product, spawns infra by terraform,
bootstraps the cluster with skuba, and configures and prepares cluster for cap.

Currently it fails on install_scf.sh, since the scripts/include/common.sh containerid/ip are not correctly set up.

For moving forward we would need to refactor the rules into steps that are k8s platform agnostic.
This can be achieved by saving the k8s external info (external ip, nfs server, etc) on a configmap on the cluster, see example at the end of `scripts/caasp4_deploy.sh`, and thinning `scripts/include/common.sh` and creating `scripts/include/<platform>.sh`

This would mean a change of the structure of the repo.

### DIFF
--- a/scripts/tools.sh
+++ b/scripts/tools.sh
@@ -14,4 +14,11 @@ if [ -z "$EKCP_HOST" ]; then
     mv kind-linux-amd64 kind
     chmod +x kind
 fi
+
+cat <<HEREDOC > .envrc
+export KUBECONFIG=$(pwd)/kubeconfig
+export HELM_HOME=$(pwd)/.helm
+export CF_HOME=$(pwd)/.cf
+HEREDOC
+
 popd
--- a/scripts/tools.sh
+++ b/scripts/tools.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
 set -x
 
-mkdir build${CLUSTER_NAME}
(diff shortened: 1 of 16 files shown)


scripts/tools.sh
```diff
@@ -10,4 +14,11 @@ if [ -z "$EKCP_HOST" ]; then
     mv kind-linux-amd64 kind
     chmod +x kind
 fi
-popd
\ No newline at end of file
+
+cat <<HEREDOC > .envrc
```
mudler: Nice :100:

mudler:
> For moving forward we would need to refactor the rules into steps that are k8s platform agnostic.

Sure, makes perfectly sense. I would just keep the simplest use case as default (so we keep the `local` approach as the one expected with plain default usage)

> This can be achieved by saving the k8s external info (external ip, nfs server, etc) on a configmap on the cluster, see example at the end of `scripts/caasp4_deploy.sh`, and thinning `scripts/include/common.sh` and creating `scripts/include/<platform>.sh`
> 
> This would mean a change of the structure of the repo.

Sounds good to me!

PR #7, opened 2019-09-03
Various fixes

jimmykarily:


scripts/include/common.sh
```diff
@@ -6,23 +6,28 @@ if [ -n "$VALUES_OVERRIDE" ] && [ -f "$VALUES_OVERRIDE" ]; then
   OVERRIDE=$(cat $VALUES_OVERRIDE)
 fi
 
-export cluster_name=${CLUSTER_NAME:-kind}
-export BUILD_DIR=build${cluster_name}
+export CLUSTER_NAME=${CLUSTER_NAME:-kind}
+export BUILD_DIR=build${CLUSTER_NAME}
+
+export KUBECONFIG=$PWD/$BUILD_DIR/kubeconfig
```
jimmykarily: Used absolute directory to let it work no matter what the current directory is in each script

scripts/include/common.sh
```diff
@@ -36,10 +39,12 @@ set +x
 export DOCKER_USERNAME="${DOCKER_USERNAME:-}"
 export DOCKER_PASSWORD="${DOCKER_PASSWORD:-}"
 export CLUSTER_PASSWORD="${CLUSTER_PASSWORD:-password}"
-set -x
+
```
jimmykarily: I can put this back. I removed this because we assume output was on so after sourcing this file we always get the verbose output.

scripts/include/common.sh
```diff
@@ -14,15 +14,19 @@ export BUILD_DIR=build${CLUSTER_NAME}
 
 export ROOT_DIR="$(git rev-parse --show-toplevel)"
 export CHART_URL="${CHART_URL:-}"
-export KUBECONFIG=$PWD/kubeconfig
+export KUBECONFIG=$BUILD_DIR/kubeconfig
```
mudler: Mmm in this way it won't be anymore an absolute path, right? but afaict shouldn't be an issue (the only thing is it would allow `kubectl` to be called inside the make targets also if there are any pushd/popd around)

scripts/include/common.sh
```diff
@@ -14,15 +14,19 @@ export BUILD_DIR=build${CLUSTER_NAME}
...
 export SCF_BRANCH="${SCF_BRANCH:-develop}"
 if [ -n "$EKCP_HOST" ]; then
   export container_ip=$(curl -s http://$EKCP_HOST/ | jq .ClusterIPs.${CLUSTER_NAME} -r)
   export DOMAIN="${CLUSTER_NAME}.${container_ip}.${EKCP_DOMAIN}"
 else
-  export container_id=$(docker ps -f "name=${cluster_name}-control-plane" -q)
-  export container_ip=$(docker inspect $container_id | jq -r .[0].NetworkSettings.Networks.bridge.IPAddress)
+  echo "CLUSTER_NAME is ${CLUSTER_NAME}"
+  export container_id=$(docker ps -f "name=${CLUSTER_NAME}-control-plane" -q)
+  if [ ! -z $container_id ]; then
```
mudler: :100:

scripts/include/common.sh
mudler: nvm, didn't saw the commit rightafter

PR #10, opened 2019-09-04
support different OS hosts when downloading the kind file

### LINKED ISSUE #9: Identify the OS Host Type
**description**

when downloading the kind file, the script always selects the linux file independently of the OS host being a mac (darwin) and that breaks the make flow.

**solution**

1. add a _if-then_ clause inside of the _common.sh_ that stores the OS host type for later usage 
`
if [[ "$OSTYPE" == "darwin"* ]]; then
  export KIND_OS_TYPE="${KIND_OS_TYPE:-kind-darwin-amd64}"
else
  export KIND_OS_TYPE="${KIND_OS_TYPE:-kind-linux-amd64}"
fi
`

2. change the tools.sh to support the new env var
`wget https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/${KIND_OS_TYPE}`

fargozhu:
Fixes #9

### DIFF
--- a/scripts/include/common.sh
+++ b/scripts/include/common.sh
@@ -31,12 +31,6 @@ fi
 
 export DEEP_CLEAN="${DEEP_CLEAN:-false}" # If true, triggers helm to delete releases before cleaning up
 export KIND_VERSION="${KIND_VERSION:-0.2.1}"
-if [[ "$OSTYPE" == "darwin"* ]]; then 
-  export KIND_OS_TYPE="${KIND_OS_TYPE:-kind-darwin-amd64}" 
-else 
-  export KIND_OS_TYPE="${KIND_OS_TYPE:-kind-linux-amd64}" 
-fi
-
 export HA="${HA:-false}"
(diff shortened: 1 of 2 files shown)


scripts/include/common.sh
```diff
@@ -31,6 +31,12 @@ fi
 
 export DEEP_CLEAN="${DEEP_CLEAN:-false}" # If true, triggers helm to delete releases before cleaning up
 export KIND_VERSION="${KIND_VERSION:-0.2.1}"
+if [[ "$OSTYPE" == "darwin"* ]]; then 
+  export KIND_OS_TYPE="${KIND_OS_TYPE:-kind-darwin-amd64}" 
+else 
+  export KIND_OS_TYPE="${KIND_OS_TYPE:-kind-linux-amd64}" 
+fi
+
```
jimmykarily: I think it's better if you moved these to the file below (tools.sh). They are not needed anywhere else after all.

PR #12, opened 2019-09-04
Refactor scripts to provide defined states between targets

viccuad:
- Use .envrc for each script that acts on a deployment, mimicking the local user, simplifying scripts and making sure that helm,kubeconfig,cf environments are different when having multiple deployments. This also allows easy cleaning of scf, helm, charts, etc, which in turn allows easy reuse of long-running k8s clusters.
- Save external k8s cluster info (domain, external ip, etc) in a configmap when deploying k8s clusters, use it in following steps.
- Rename and reorder scripts. 

No externally used make targets have been renamed; all pipelines should continue to work.

---
## States

This refactor aims to enable the following states:

### 1. K8s deploy and prepare

Deploys whatever kubernetes, prepares it for CAP adding a storageclass, RBAC,
helm, and adding a configmap to the k8s cluster containing outside information
(external ip, domain, filesystem of the nodes, etc).

Output: a buildir containing kubeconfig, HELM_HOME initialized by helm, .envrc
so one can enter each and every different k8s cluster indepently of one another
and the bkindwscf host.

### 2. CAP deploy

Takes whatever k8s deployment prepared for CAP and generates scf-config-values, obtains charts, installs scf
and logins with cf-cli.

Output: a buildir of a deployment with SCF installed, CF_HOME configured.

Can be undone by `make clean-scf`

### 2.1. Stratos deploy

Takes previous state and install stratos.

### 3. Test CAP

Takes a state 2 and runs CAPS tests on it.

Output: saves test results at `build/artifacts/*`


viccuad:
The last 2 commits renaming `scripts/*` and reordering `Makefile` could be dropped.

viccuad:
travis should have failed, they do `make kind gen-config setup`.

This refactor needs `make kind setup chart gen-config`; setup creates the k8s configmap, gen-config reads vars from the k8s configmap. chart creates the chart folder, gen-config reads the charts for the DEFAULT_STACK if chart folder exists.

scripts/scf_install.sh
```diff
@@ -3,13 +3,21 @@
 set -ex 
 
 . scripts/include/common.sh
+. .envrc
 
 
+if [[ $ENABLE_EIRINI == true ]] ; then
+    kubectl create namespace eirini
+    helm install stable/metrics-server --name=metrics-server \
```
mudler: :100:

scripts/kind_up.sh
```diff
@@ -23,4 +23,5 @@ if [ -n "$EKCP_HOST" ]; then
     curl -d "name=${CLUSTER_NAME}" -X POST http://$EKCP_HOST/new
 else
     ./kind create cluster --config kind-config.yaml --name=${CLUSTER_NAME}
+    cp $(./kind get kubeconfig-path --name="$CLUSTER_NAME") kubeconfig
```
mudler: Are we dropping the `kubeconfig` target ? mind that this mechanism is also used by the `dind` one (bootstrap SCF within the docker image)

mudler: This seems a dup, it would collide with the `make kubeconfig` target that does the same (just chain it where you need it)

Makefile
```diff
@@ -121,13 +136,20 @@ caasp4os-deploy:
...
-	scripts/eirini-release.sh
+# pipeline targets:
+
+.PHONY: kind
+kind: clean deps up kubeconfig
+
+.PHONY: all
+all: kind setup chart gen-config scf login
+
+.PHONY: dind
```
mudler: That's not used by any pipeline (but is the one that must be used when running from the docker image)

Makefile
```diff
@@ -1,79 +1,84 @@
-.PHONY: setup
-setup:
-	scripts/setup.sh
+.PHONY: buildir
+buildir:
```
mudler: :+1:

scripts/ingress_forward.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -ex 
+
+. scripts/include/common.sh
+. .envrc
+
+exec kubectl port-forward -n default pod/socksproxy 2224:8000
```
mudler: :smiley:  Mind to add a default value for 2224 and make it overridable? Was about to do it but then I would collide with your changes

scripts/scf_gen_config.sh
```diff
@@ -2,11 +2,20 @@
 set -ex
 
 . scripts/include/common.sh
+. .envrc
 
 if [ -z "${DEFAULT_STACK}" ]; then
     export DEFAULT_STACK=$(helm inspect helm/cf/ | grep DEFAULT_STACK | sed  's~DEFAULT_STACK:~~g' | sed 's~"~~g' | sed 's~\s~~g')
 fi
 
+domain=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["domain"]')
```
mudler: Smart approach! :+1:

scripts/kind_up.sh
mudler: Thinking twice: we could drop the `docker-kubeconfig` target by fusing it with `kubeconfig`, in that way we can drop also the `dind` target - they just need a switch (maybe an ENV var?) that can be pinned as default from the Dockerfile

scripts/ingress_forward.sh
mudler: done in https://github.com/SUSE/bkindwscf/commit/03e9467e0fcaff375e6da97cfb59332aa7e9840c, nevermind!

mudler:
Sorry for the merge conflicts :(

Makefile
viccuad: renamed the comment from "pipeline targets" to "full targets"

scripts/kind_up.sh
viccuad: In my mind, the canonical place to obtain a kubeconfig is `buildir/kubeconfig`; either you have it there because the k8s cluster deployment succeeded (be it kind, caasp4, caasp3, etc), or you don't have it.

Therefore I would remove the `make kubeconfig`, and keep the `make docker-kubeconfig` to be used with dind.

I can see the benefit of moving into the future to a makefile that triggers different deployments and make targets depending on an env var PLATFORM={kind,dind,caasp4os,caasp4bare,gke,eks} etc.

viccuad:
rebased onto current master :)

viccuad:
which pipeline is the one that failed? I cannot see the same error when deploying metrics or scf on my `make all` (kind, eirini, develop).

mudler:
https://concourse.suse.de/teams/main/pipelines/eirini-ci which is in https://github.com/SUSE/cloudfoundry/tree/master/ci/pipelines/eirini-ci

scripts/kind_up.sh
mudler: Just please mind https://github.com/SUSE/bkindwscf/commit/8fa83463d628397ce98961eea46ce3f82a1b5ea3 while doing this, we found other "ways" to get a kubeconfig dynamically :)

mudler:
https://concourse.suse.de/teams/main/pipelines/eirini-ci/jobs/test-eirini-develop/builds/8

scripts/kind_up.sh
viccuad: Removed the kubeconfig acquisition from scf_install. Since `make kind` and `make dind` keep having a `kubeconfig` target, everything is as it was before

PR #15, opened 2019-09-09
Typos Fixed

fargozhu:
1. the command _mv_ was using the _linux_ hardcoded name of **kind**, which in case the downloaded file was for another OS, like _darwin_ the command would fail. The fix uses the **KIND_OS_TYPE** variable that establishes the kind file to download based on host OS.

2. the **BUILD_DIR** was being created by the _buildir_ script and so it was duplicated in here.

### DIFF
--- a/scripts/kind_tools.sh
+++ b/scripts/kind_tools.sh
@@ -5,8 +5,6 @@ set -x
 export CLUSTER_NAME=${CLUSTER_NAME:-kind}
 export BUILD_DIR=build${CLUSTER_NAME}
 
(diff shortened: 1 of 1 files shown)


scripts/kind_tools.sh
```diff
@@ -5,8 +5,6 @@ set -x
 export CLUSTER_NAME=${CLUSTER_NAME:-kind}
```
viccuad: these 2 lines could also go

viccuad: feel free to merge without these

PR #19, opened 2019-09-18
Add smoke-scf target that runs smoke tests from built images on charts

viccuad:


mudler:
We should probably have an entry in the README for this, or either create the wiki here to collect such use cases

PR #20, opened 2019-09-19
Add minikube as k8s backend

viccuad:


scripts/minikube_up.sh
```diff
@@ -0,0 +1,26 @@
...
+
+. scripts/include/common.sh
+. .envrc
+
+set -Eeuxo pipefail
+
+# minikube uses retcodes to codify in binary the status of minikube, cluster, and kubernetes
+# Eg: 7 meaning: 1 (for minikube NOK) + 2 (for cluster NOK) + 4 (for kubernetes NOK)
+set +x
+minikube delete || true
```
mudler: Mmm not sure if this is a leftover, why deleting in the up script? We could do that the minikube target calls clean upfront (like `kind` does, so behaves the same)

mudler:
Nice! :100: this deserves a readme update :smile:

viccuad:
totally true, added! :)

PR #21, opened 2019-09-20
Fix minikube

viccuad:


scripts/minikube_prepare.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+set -x
+
+. scripts/include/common.sh
+. .envrc
+
+kubectl create clusterrolebinding admin --clusterrole=cluster-admin --user=system:serviceaccount:kube-system:default
```
mudler: those dups will need to go in some other common part with kind, wdyt? or why we cannot reuse scf_setup? anyway shouldn't block this PR

We could also start to think to a `BACKEND` var to switch among them (minikube, caasp, kind, ekcp, etc.) but it is out of scope now

viccuad: You are right, either move the clusterrolebindings to scf_setup, or start putting things in include. But the rest of the things (helm, tiller, storageclass) I prefer them with the k8s backend.

I would like to define the states before though, even if I have just muddled the thing right now

PR #22, opened 2019-09-23
Add make scf-precond to check k8s cluster

viccuad:
Performs `kube-ready-state-check.sh kube` only, not all, as we cannot have
access to the physical nodes.


mudler:
Maybe `scf-check` or even `check`? (`precond` sounds arcane! :smile: )

viccuad:
True! Changed it to `scf-precheck`, for people that don't know about scf I think it may be more fitting that `make check` which may sound like running tests against an already deployed scf.

Apart from updating the wiki, I would like to have an `scf` target that completely prechecks, gets charts, deploys, etc. But that would mean that pipelines would need to be changed.

PR #24, opened 2019-09-26
Fix chart in capvalues

viccuad:
Add CHART_URL at scf_install and remove it when scf_clean

### DIFF
--- a/scripts/scf_install.sh
+++ b/scripts/scf_install.sh
@@ -5,6 +5,9 @@ set -ex
 . scripts/include/common.sh
 . .envrc
 
+# save CHART_URL on cap-values configmap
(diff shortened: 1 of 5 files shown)


scripts/scf_install.sh
```diff
@@ -5,6 +5,9 @@ set -ex
 . scripts/include/common.sh
 . .envrc
 
+# save CHART_URL on cap-values configmap
+kubectl patch -n kube-system configmap cap-values -p $'data:\n chart: "$CHART_URL"'
```
mudler: Probably the same should happen during upgrade

viccuad: done in https://github.com/SUSE/bkindwscf/pull/25

PR #28, opened 2019-09-27
Remove $CF_HOME/.cf, not $CF_HOME, which points to buildir

viccuad:
On `make clean-scf` we need to delete all HOMEs, including CF_HOME. While it's
true that CF_HOME defaults to ~, and cf looks for files in CF_HOME/.cf, if we
leave CF_HOME="$(pwd)" it points to the current dir and we cannot delete it.

If we don't fix this, on `make clean-scf`, the buildir gets deleted.

Let the cfcli config files be in buildir/.cf/cf

### DIFF
--- a/scripts/scf_clean.sh
+++ b/scripts/scf_clean.sh
@@ -28,7 +28,7 @@ if [[ $ENABLE_EIRINI == true ]] ; then
     helm del --purge metrics-server
 fi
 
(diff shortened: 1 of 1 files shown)


jimmykarily:
I think it is better is we replace this line https://github.com/SUSE/catapult/blob/master/scripts/scf_clean.sh#L31
 with

```
rm -rf scf-config-values.yaml chart.zip helm kube "$HELM_HOME" "${CF_HOME}/.cf"
```

viccuad:
Doh, done.

PR #29, opened 2019-09-30
Add `make gke`, `make clean-gke`, and `make all-gke` targets

viccuad:


mudler:
Maybe a mention in the wiki about gke? :100:

viccuad:
Tested  `make all-gke`, added a wait_ns for the k8s deployment, made installing metrics-server on scf_install.sh idempotent.

I failed on the eirini deployment for eirini-copy-certs-x pod as I forgot to pass the registry env vars, but I think it's good to go.

Issue #38, opened 2019-10-02
Split unit tests in its own subfolder


mudler:
This script https://github.com/SUSE/catapult/blob/master/scripts/tests_catapult.sh is exercising catapult internally for the config generation (can be triggered with `make catapult-test`). 

We should move those bits to a separate subfolder and organize properly the testcases.

mudler:
done in https://github.com/SUSE/catapult/pull/48

Issue #35, opened 2019-10-01
Simplify layout of scripts/*


viccuad:
Currently, the implementation consists of everything flat on `scripts/*`. Add folders per k8s backend, state, etc (`kind/*` ,`caasp4os/*` , `gke/*` , `scf/*` , `tests/*`). It would be nice to document idiosyncrasies or what each state consists of, in wiki, readme files in the folder, etc.

viccuad:
Ongoing work of a possible reimplementation  at `v2-layout` branch, comments welcome

mudler:
> Ongoing work of a possible reimplementation at `v2-layout` branch, comments welcome

Had a quick look. I would invest a bit into porting to functions some of the recurring patterns, e.g. extracting information from configmap (our `cap-values`), wait for namespace, etc.. they should be treated as helpers

viccuad:
I will add topical folders and change the makefile, and open a PR to merge against a `v2` branch. I'm scared of trying to change everything at once and be a bottleneck for the refactoring efforts

mudler:
Closing, fixed in https://github.com/SUSE/catapult/pull/48

Issue #34, opened 2019-10-01
Split extensions from core functionalities


mudler:
So the make targets can be more clear on what they do.

#31

mudler:
Done in https://github.com/SUSE/catapult/pull/48

Issue #33, opened 2019-10-01
More transparent and clear the make tagets scopes


mudler:
**Problem:** There is no real differentiation between private and public targets, and backend-private targets are currently used for end-user usage.

We should split make targets between `internal` and `public` so to provide a stable and solid API outside the internal usage. 

This also means that we should document the public one, providing also a description of the arguments that can tweak the targets scopes.

E.g:

Internal:

- `make minikube`
- `make kind`
- `make kind-clean`
- `make minikube-clean`

Public:

- `make cluster` / `make kubernetes`
- `make scf`
- `make clean`
- `make scf-login`
- `make all`

Public (Extra-core functions, Plugins):

- `make ingress` and `make ingress-forward`
- `make terminal`
...

In such way we can define a configuration aka "Context" which our make targets run against. A Context can be a JSON/YAML (https://github.com/SUSE/catapult/issues/30) or a ENV configuration (e.g. `CLUSTER_NAME=gke-test1`, `BACKEND=gke` )

Relates to: https://github.com/SUSE/catapult/issues/40 and https://github.com/SUSE/catapult/issues/34

fargozhu:
@mudler can u describe a little bit more what you want to achieve?

viccuad:
Closing this as everything was already implemented :tada:

PR #46, opened 2019-10-02
Change layout to script/<folder> and makefiles in each folder

viccuad:
As a first step, this changes the layout to one where topical things are in scripts/<folder>. Each folder has its on Makefile.

Paths have been transformed to absolute paths when possible using `$ROOT_DIR`.

I haven't tested all the make targets though.


viccuad:
I have squashed commits as sanely as possible, this partially implements https://github.com/SUSE/catapult/issues/35.

No effort has been done yet in extracting recurring patterns into functions to be included at `scripts/include/*`, but I would be glad to merge this after changing travis to make it work.

scripts/tools/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+.PHONY: ktw
+kwt:
+	./kwt.sh
+
+.PHONY: kwt-connect
+kwt-connect:
+	./kwt_connect.sh
+
+.PHONY: task
```
mudler: I wouldn't put task/terminal/web to tools, but instead to something like `extra` or `plugins` (as they are internal, no external tooling provided)

scripts/unit/Makefile
```diff
@@ -0,0 +1,5 @@
+.DEFAULT_GOAL := tests_catapult.sh
+
+.PHONY: tests-catapult
```
mudler: `unit` is too much scoped imho, we might end up to put also some integration tests at some point. catapult tests problably should be outside of `scripts/` and just be a top level folder

mudler:
Overall it looks really good! I would even merge to master eventually because doesn't seem to be harmful and we could instead incrementally work on the v2 (and no need of separate branch).

Tests are failing somehow, seems make targets broken.

scripts/ingress/Makefile
```diff
@@ -0,0 +1,10 @@
+# no default goal
+# .DEFAULT_GOAL :=
```
mudler: Also this should be part of `plugins` or similar

viccuad:
unit tests and minimal kind test is passing :). I definitely could rebase and squash some commits though.

Kind deployments work, the others I have tested some, but maybe there are minor problems here and there. I have mixed feelings on merging without testing the other deployments.

This fully implements https://github.com/SUSE/catapult/issues/38, https://github.com/SUSE/catapult/issues/35, https://github.com/SUSE/catapult/issues/34

It partially implements https://github.com/SUSE/catapult/issues/33: public targets are provided in the root Makefile, accesible with `make ^TAB`. Private targets are inside of the scripts/<topic>.

It partially implements https://github.com/SUSE/catapult/issues/32: `make scf` does what `make chart gen-config scf login` used to do.

mudler:
> unit tests and minimal kind test is passing :). I definitely could rebase and squash some commits though.

Nice! Rebasing/squashing would definetly ease a bit the review, but as you  prefer :+1: 
> 
> Kind deployments work, the others I have tested some, but maybe there are minor problems here and there. I have mixed feelings on merging to master without testing the other deployments, imho we should merge to v2, and keep working there.

I can point the CI on this branch :)

> 
> This fully implements #38, #35, #34
> 
> It partially implements #33: public targets are provided in the root Makefile, accesible with `make ^TAB`. Private targets are inside of the `scripts/<topic>`, so one needs to `make -C scripts/<topic> ^TAB` or `cd` into them for now.
> 
> It partially implements #32: `make scf` does what `make chart gen-config scf login` used to do. If you want to se the old targets, you will need `make -C scripts/scf chart gen-config scf login` or a subset of those.
> Changing the `scf` target means that merging this will break existing pipelines.

What about having an `intermediate` make target which we would deprecate later to support legacy and avoid breakage?

tests/tests_catapult.sh
```diff
@@ -0,0 +1,48 @@
...
+  assertContains 'contains CF_HOME' "$ENVRC" 'CF_HOME="$(pwd)"'
+  make clean
+  assertTrue 'clean buildir' "[ ! -d 'buildtest' ]"
+}
+
+# Tests values generation
+testConfig() {
+  rm -rf buildtest
+  make buildir
+  DEFAULT_STACK=sle CLUSTER_PASSWORD=test123 make -C scripts/scf gen-config
```
mudler: Why no gen-config on the top-level make target ? Imho even if it should be treated as internal we still should have a bridge(or whatever) that allows us to call such targets with more ease. Even a simple "agreement" like: all internal targets should be in form `private-(module)-(action)`, but that's just my 2c

viccuad: That could be definitely done, haven't made more internal api separation than the makefiles inside of the folders, which I think is useful for newcomers doing just `make ^TAB`.
I'm concerned about adding more things to this PR, I would like to merge it asao to reduce the bottleneck/rebases

viccuad:
> I can point the CI on this branch :) If CI is good, after review we could directly merge on master imho

let me rebase and squash when I have time, and then we can try to merge. Soo looking forward to merging.

> What about having an intermediate make target which we would deprecate later to support legacy and avoid breakage?

I could add a public `make deploy-scf` or the like that encompasses all the scf targets, and expose the current ones as `scf-*`, but imho it is a good moment to simplify the public api. Honest question, do we normally use the scf targets on a 1 by 1 basis?

viccuad:
rebased onto master, and squashed commits. I'm happy merging this to master :relaxed:

tests/tests_catapult.sh
mudler: I know.. and do not want to push this much over, I will try to add few commits on top of this work and give it a try on the CI

PR #49, opened 2019-10-14
[ci test] Add catapult integration tests

mudler:
~~this should trigger on travis a full SCFv3 deployments + smoke passing.~~ It will add integration tests

~~with SCFv2 wasn't possible for the deployment times which were exceeding travis limits, let's try now with SCFv3~~ we exhaust the limits, it is still not possible

### DIFF
--- a/tests/tests_catapult_integration.sh
+++ b/tests/tests_catapult_integration.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+ROOT_DIR="$(git rev-parse --show-toplevel)"
+export CLUSTER_NAME=test
+export BACKEND=kind
+export ENABLE_EIRINI=false
+
+[ ! -d "shunit2" ] && git clone https://github.com/kward/shunit2.git
+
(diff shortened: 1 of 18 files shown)


tests/tests_catapult_integration.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+ROOT_DIR="$(git rev-parse --show-toplevel)"
+export CLUSTER_NAME=test
+export SCF_OPERATOR=true
+export BACKEND=kind
+export ENABLE_EIRINI=false
+export CHART_URL=https://scf-v3.s3.amazonaws.com/scf-3.0.0-380c355.tgz
```
viccuad: Would it be possibleto  use a symlink to a working scf v3 on s3 to not hardcode a chart?

mudler: It exhaust the travis limits anyway, so we can't do this :(

PR #56, opened 2019-10-15
Mark deprecated targets as deprecated

viccuad:
Wait until user presses a key before continuing or they don't see anything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ clean:
 	make -C backend/$(BACKEND) clean
 
 .PHONY: k8s
-k8s:
+k8s: clean buildir
 	make -C backend/$(BACKEND)
 
 .PHONY: kubeconfig
--- a/Makefile
+++ b/Makefile
@@ -204,67 +204,67 @@ sample-ticking:
 
 .PHONY: kind
 kind::
-	@echo 'This target is deprecated. Please use `make k8s` or `BACKEND=kind make k8s` instead'
-	@read -p 'Press any key to continue' any
+	@echo 'WARNING: target deprecated. Please use `make k8s` or `BACKEND=kind make k8s` instead.'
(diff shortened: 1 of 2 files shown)


viccuad:
Heh, travis ci timed out because of the `Press any key to continue`. Should we use a sleep instead, so the user gets a bit more time to read the deprecation notice, but things keep running?

mudler:
A sleep to give time to read sounds better so we don't break any automation already in place ( and at the same time we keep Travis happy )

viccuad:
I think something is broken between BACKEND=, CLUSTER_NAME=, and the lack of them when building BUILD_DIR.

Makefile
```diff
@@ -19,6 +19,10 @@ buildir:
 clean:
 	make -C backend/$(BACKEND) clean
 
+.PHONY: k8s
+k8s:
```
mudler: It is breaking because k8s should call `clean buildir` first (as the deprecated targets are doing)

tests/tests_catapult_integration.sh
```diff
@@ -42,7 +42,7 @@ tearDown() {
 
 testKind() {
     rm -rf buildtest
-    make kind
+    BACKEND=kind make k8s
```
mudler: no buildir is created by the target, that's why travis complains with:
`mv: cannot move 'kind' to 'bin/kind': No such file or directory

Makefile:9: recipe for target 'deps' failed

make[3]: *** [deps] Error 1

make[3]: Leaving directory '/home/travis/build/SUSE/catapult/backend/kind'

Makefile:24: recipe for target 'k8s' failed`

Makefile
viccuad: Wooops. That's embarrashing. Thanks!

PR #59, opened 2019-10-16
Add `make stratos-upgrade`

viccuad:
This PR adds a public `make stratos-upgrade`. Also:

- Pass stratos chart on `STRATOS_CHART`
- Deprecate `CHART_URL`, add `SCF_CHART` that takes the previous.
- Add private `make -C modules/stratos chart` to put the chart on `buildir/console`
- Add private `make -C modules/stratos gen-config` to copy the scf-config-values.yaml to scf-config-values-for-stratos.yaml, where one can point to a different registry.

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -19,6 +19,7 @@ export BUILD_DIR="$ROOT_DIR"/build${CLUSTER_NAME}
 [ -d "$BUILD_DIR" ] && pushd "$BUILD_DIR"
 
 export CHART_URL="${CHART_URL:-}"
+export STRATOS_CHART="${STRATOS_CHART:-}"
 export SCF_REPO="${SCF_REPO:-https://github.com/SUSE/scf}"
 export SCF_BRANCH="${SCF_BRANCH:-develop}"
 
--- a/include/common.sh
+++ b/include/common.sh
@@ -18,7 +18,8 @@ export BUILD_DIR="$ROOT_DIR"/build${CLUSTER_NAME}
 # Forces our build context
 [ -d "$BUILD_DIR" ] && pushd "$BUILD_DIR"
 
-export CHART_URL="${CHART_URL:-}"
+export CHART_URL="${CHART_URL:-}" # FIXME deprecated, used in SCF_CHART
+export SCF_CHART="${SCF_CHART:-CHART_URL}"
 export STRATOS_CHART="${STRATOS_CHART:-}"
 export SCF_REPO="${SCF_REPO:-https://github.com/SUSE/scf}"
 export SCF_BRANCH="${SCF_BRANCH:-develop}"
--- a/modules/scf/Makefile
+++ b/modules/scf/Makefile
@@ -38,7 +38,7 @@ scf-brats-setup:
 	./brats_setup.sh
 
 .PHONY: upgrade
-upgrade:
+upgrade: chart gen-config
 	./upgrade.sh
 
 .PHONY: build-scf-from-source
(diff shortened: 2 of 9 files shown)


include/common.sh
```diff
@@ -18,7 +18,9 @@ export BUILD_DIR="$ROOT_DIR"/build${CLUSTER_NAME}
 # Forces our build context
 [ -d "$BUILD_DIR" ] && pushd "$BUILD_DIR"
 
-export CHART_URL="${CHART_URL:-}"
+export CHART_URL="${CHART_URL:-}" # FIXME deprecated, used in SCF_CHART
+export SCF_CHART="${SCF_CHART:-CHART_URL}"
```
mudler:
```make[3]: Leaving directory '/home/travis/build/SUSE/catapult/modules/scf'

Makefile:176: recipe for target 'scf-chart' failed

make[2]: *** [scf-chart] Error 2

make[2]: Leaving directory '/home/travis/build/SUSE/catapult'

ASSERT:helm folder is present```

I think is caused because it should be: ` export SCF_CHART="${SCF_CHART:-$CHART_URL}"`

File: modules/scf/clean.sh [hunk 2]
```diff
@@ -32,7 +32,7 @@ if [[ "$ENABLE_EIRINI" == true ]] ; then
     fi
 fi
 
-rm -rf scf-config-values.yaml chart.zip helm kube "$CF_HOME"/.cf
+rm -rf scf-config-values.yaml chart helm kube "$CF_HOME"/.cf
```
mudler: Good catch! :)

modules/stratos/chart.sh
```diff
@@ -0,0 +1,30 @@
...
+
+. ../../include/common.sh
+. .envrc
+
+set -Eexuo pipefail
+debug_mode
+
+if [ -z "$STRATOS_CHART" ]; then
+    warn "No chart url given - using latest public release from GH"
+        STRATOS_CHART=$(curl -s https://api.github.com/repos/cloudfoundry/stratos/releases/latest | grep "browser_download_url.*zip" | cut -d : -f 2,3 | tr -d \" | tr -d " ")
```
mudler: really nice! Probably we should move the gh fetch to its own function at some point

mudler: (or even the whole chart fetch/uncompress)

viccuad: well that's entirely lifted from scf-chart :D . It works though.

include/common.sh
viccuad: toootally true! thanks! :D

viccuad:
This is missing renaming CHART_URL on the rest of the targets, too.

mudler:
> This is missing renaming CHART_URL on the rest of the targets, too.

This is fine, we can create a separate issue and adapt later

modules/scf/Makefile
```diff
@@ -38,7 +38,7 @@ scf-brats-setup:
 	./brats_setup.sh
 
 .PHONY: upgrade
-upgrade:
+upgrade: chart gen-config
```
mudler: Just noticed, this also breaks other workflow, IMHO upgrade should just upgrade, and do anything else. Maybe move this to a separate private target?

viccuad: I really wonder, because upgrading will by definition always use a different chart and/or registry creds than the install already present, so we need to recompute both just in case people are passing the options.
I'm thinking, but can't find another workflow right now.

mudler: well, one that comes up to me is: one can deploy, and right after call upgrade to add an env var by modifying the values file manually, no need to uncompress the chart or regenerate the config in such cases.

PR #62, opened 2019-10-16
Add `make stratos-upgrade` redux

viccuad:
This PR adds a public make stratos-upgrade. Also:
- Pass stratos chart on STRATOS_CHART
- Deprecate CHART_URL, add SCF_CHART that takes the previous.
- Add private make -C modules/stratos chart to put the chart on buildir/console
- Add private make -C modules/stratos gen-config to copy the scf-config-values.yaml to scf-config-values-for-stratos.yaml, where one can point to a different registry.

### DIFF
--- a/modules/scf/Makefile
+++ b/modules/scf/Makefile
@@ -38,7 +38,7 @@ scf-brats-setup:
 	./brats_setup.sh
 
 .PHONY: upgrade
-upgrade: chart gen-config
+upgrade:
 	./upgrade.sh
 
 .PHONY: build-scf-from-source
--- a/modules/scf/Makefile
+++ b/modules/scf/Makefile
@@ -38,7 +38,7 @@ scf-brats-setup:
 	./brats_setup.sh
 
 .PHONY: upgrade
-upgrade:
+upgrade: chart gen-config
 	./upgrade.sh
 
 .PHONY: build-scf-from-source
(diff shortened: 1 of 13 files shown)


viccuad:
Missing resolution of: https://github.com/SUSE/catapult/pull/59#discussion_r335492162

modules/scf/Makefile
```diff
@@ -38,7 +38,7 @@ scf-brats-setup:
 	./brats_setup.sh
 
 .PHONY: upgrade
-upgrade:
+upgrade: chart gen-config
```
mudler: Shall we have a helm-release target for upgrades without a chart ? (e.g. env changes, settings, etc) 
This needs doc updates https://github.com/SUSE/catapult/wiki/Running-options#upgrading-a-deployment

viccuad: I have reverted the changes to `scf-upgrade`, feels like complex enough to discuss on an issue.
I'm keeping `stratos-upgrade` as it is, it depends on `stratos-chart` and `stratos-gen-config`, so one needs to pass all env vars that one passed when deploying stratos, plus the changes that one wants.

viccuad: And by starting to write an issue on the topic, I have backpedaled and made it so one would need to do `OPTS='' make stratos-chart stratos-gen-config stratos-upgrade`, I'm a convert :stuck_out_tongue_closed_eyes:

Issue #55, opened 2019-10-15
Add make target for importing existing clusters


viccuad:
Currently, if one wants to run catapult make targets against a preexisting target, one needs to do
`CLUSTER_NAME=foo make buildir; cp kubeconfig buildir/`, and make sure that the `cap-values` configmap exists. (see https://github.com/SUSE/catapult/wiki/Import-your-k8s-cluster).

We could streamline this by providing a make target. It may be useful to detect on the target if the cluster is prepared for cap by running the kube-check script from github.com/SUSE/scf, or interactively prompting for values to create a cap-values configmap.

mudler:
I would really like that!

viccuad:
This seems to be more difficult than it looks at first view,

- Needs to work when sourcing the script, and when `direnv` "loads" the script (spawns a bash subshell, executes the shell, captures the difference in env vars, exports the env vars outside).
- Should not use readlink, as for eg. macos may not have it installed by default
- We could use $BASH_SOURCE, but that's only present on bash v3+ (again, macos)
- Should work if sourced from zsh or other shells

viccuad:
Implemented in https://github.com/SUSE/catapult/pull/63

PR #63, opened 2019-10-16
Add backend/imported/* and $KUBECONFIG to pass a config to it

viccuad:
Usage:
  BACKEND=imported KUBECONFIG=<abs path> make k8s
  BACKEND=imported make clean

### DIFF
--- a/backend/imported/Makefile
+++ b/backend/imported/Makefile
@@ -1,0 +1,20 @@
+.DEFAULT_GOAL := all
+
+.PHONY: kubeconfig
+kubeconfig:
+	./kubeconfig.sh
+
+.PHONY: clean
+clean:
+	./clean.sh
+
+.PHONY: deps
+deps:
+	./deps.sh
+
+.PHONY: check
+check:
+	./check.sh
+
+.PHONY: all
+all: deps kubeconfig check
--- a/backend/imported/check.sh
+++ b/backend/imported/check.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+. ../../include/common.sh
+. .envrc
+
+set -Eeuxo pipefail
+debug_mode
+
+curl -Lo kube-ready-state-check.sh "$SCF_REPO"/raw/"$SCF_BRANCH"/bin/dev/kube-ready-state-check.sh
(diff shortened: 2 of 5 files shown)


viccuad:
Right now cap-values contains the ip of the first worker node, and domain as omg.h.w. We could interactively ask for a different domain.

mudler:
Very nice!  Care to add a simple test ? :smiley_cat: 

> Right now cap-values contains the ip of the first worker node, and domain as omg.h.w. We could interactively ask for a different domain.

I would keep the interaction low, but afaik ekcp has already a `EKCP_DOMAIN`, probably we should have a global `DOMAIN` to refer between all of the backends (defaults to nip.io currently)

viccuad:
Would it be ok to refactor the `make catapult-test` to search recursively for files named eg `tests_catapult.sh` and execute them?

mudler:
> Would it be ok to refactor the `make catapult-test` to search recursively for files named eg `tests_catapult.sh` and execute them?

that would be nice!

include/common.sh
```diff
@@ -31,12 +31,14 @@ set +x
 export DOCKER_USERNAME="${DOCKER_USERNAME:-}"
 export DOCKER_PASSWORD="${DOCKER_PASSWORD:-}"
 export CLUSTER_PASSWORD="${CLUSTER_PASSWORD:-password}"
+set -x
```
mudler: I would rather avoid to be verbose here by default, [debug_mode](https://github.com/SUSE/catapult/blob/master/include/func.sh#L150) is called [here](https://github.com/SUSE/catapult/blob/master/include/common.sh#L6) so it will get verbose if `DEBUG_MODE` is set

viccuad: Argh, true, thanks!
The problem though is that after the `set +x` for `DOCKER_*` creds, we want to reenable if `DEBUG_MODE=true` or we don't list the rest. I will substitute it with a call to `debug_mode`.

viccuad:
At the end added testBackendImported(), not a recursive look for testfiles.

viccuad:
Closes https://github.com/SUSE/catapult/issues/55

PR #66, opened 2019-10-18
Add json config and helm repo fetch support

mudler:
It adds support to config files and allows to directly fetch and re-assemble charts from the official helm repo.

Note: the `BACKEND` cannot yet be annotated in the json as it is needed in the Makefile and it will be ignored. And ENV vars have precedence over the one specified in the json (done on purpose!).

Json config:
`CONFIG=/path/to/json BACKEND=backend make [..]`

Use chart from helm (need to pass the SCF version):
`SCF_HELM_VERSION=2.14.5 make [..]`

e.g.:
```
$> cat support.json
{
  "SCF_HELM_VERSION": "2.14.5",
  "CLUSTER_NAME": "support",
  "ENABLE_EIRINI": "false",
  "EKCP_HOST": "....
}
```

### DIFF
--- a/include/buildir.sh
+++ b/include/buildir.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 # duplicated in s/include/common.sh, needed for bootstrapping:
-. include/common.sh
+. $ROOT_DIR/include/common.sh
 
 info "Creating $BUILD_DIR"
 mkdir "$BUILD_DIR"
-. include/common.sh # Reload, as we just created BUILD_DIR
+. $ROOT_DIR/include/common.sh # Reload, as we just created BUILD_DIR
 
 mkdir bin
 
--- a/modules/scf/chart.sh
+++ b/modules/scf/chart.sh
@@ -63,11 +63,11 @@ if [ -n "$SCF_HELM_VERSION" ]; then
     popd
 fi
 
-rm -rf chart_url || true
+rm -rf scf_chart_url || true
 
 if echo "$SCF_CHART" | grep -q "http"; then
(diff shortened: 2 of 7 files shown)


modules/scf/chart.sh
```diff
@@ -63,9 +63,11 @@ if [ -n "$SCF_HELM_VERSION" ]; then
     popd
 fi
 
+rm -rf chart_url || true
+
 if echo "$SCF_CHART" | grep -q "http"; then
-    info "Downloading $SCF_CHART"
     wget "$SCF_CHART" -O chart
+    echo "$SCF_CHART" > chart_url
```
viccuad: we should differentiate it from the stratos chart, maybe `scf_chart_url`?

include/buildir.sh
```diff
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 # duplicated in s/include/common.sh, needed for bootstrapping:
-. include/common.sh
+. $ROOT_DIR/include/common.sh
```
viccuad: how does this work? isn't $ROOT_DIR undefined until one sources include/common.sh?

mudler: It's in the top level of the Makefile, so it's always present :)

modules/scf/chart.sh
mudler: Addressed in https://github.com/SUSE/catapult/pull/66/commits/1a970299781bc5b4c05f59f1f3a5ef347fb44f39

viccuad:
I love this, many thanks! :smile:

PR #68, opened 2019-10-23
Trust the Kube ca cert on the node

jimmykarily:


backend/kind/up.sh
```diff
@@ -25,4 +25,8 @@ EOF
 
 kind create cluster --config kind-config.yaml --name=${CLUSTER_NAME}
 
+# Trust the kubernetes ca on the node so CF application containers can be pulled
+# from the registry (otherwise it fails because the ca is not trusted).
+docker exec -it ${CLUSTER_NAME} bash -c 'cp /etc/kubernetes/pki/ca.crt /etc/ssl/certs/ && update-ca-certificates && systemctl restart containerd'
```
mudler: That needs to be tied to the kind versions that are using containerd, right ? but we could also make it not fail for running this line

mudler: Probably better in `prepare.sh`, because at that point we are pretty sure the k8s cluster is up ad ready

PR #70, opened 2019-10-25
 Add catapult-lint with shell and yaml checks

viccuad:
catapult-test now depends on catapult-lint, which gets run first and it's fatal.
Let's enable just errors for now, there's too many warnings…


viccuad:
So, from travis:
```
$ bash -c 'shopt -s globstar nullglob; shellcheck --severity=error **/*.{sh,ksh,bash}'

In include/func.sh line 2:

#!/bin/bash

^-- SC1128: The shebang must be on the first line. Delete blanks and move comments.

In modules/scf/purge.sh line 14:

for buildpack in $(cf buildpacks | tail -n +4 | gawk '{print $1} do'); cf delete-buildpack -f $buildpack; done

^-- SC1073: Couldn't parse this for loop. Fix to allow more checks.

                                                                       ^-- SC1058: Expected 'do'.

                                                                       ^-- SC1072: Expected 'do'. Fix any mentioned problems and try again.

For more information:

  https://www.shellcheck.net/wiki/SC1058 -- Expected 'do'.

  https://www.shellcheck.net/wiki/SC1128 -- The shebang must be on the first ...

  https://www.shellcheck.net/wiki/SC1072 -- Expected 'do'. Fix any mentioned ...

The command "bash -c 'shopt -s globstar nullglob; shellcheck --severity=error **/*.{sh,ksh,bash}'" exited with 1.
```

viccuad:
Fixed the shell errors, and now travis should pass :tada:

mudler:
Nice! :) probably we should have something like `catapult-lint` or directly embed this in `catapult-test` so one can run this locally as well ( and maybe make them fatal even? )

viccuad:
you are totally right! :). Second take, added as catapult-lint, which is fatal, and which catapult-test depends on.

viccuad:
Turns out travis' ubuntu 16.04 image doesn't have `yamllint` installed, nor has a cache populated with `apt update`, so I'm installing it by pip.

The lint.sh fails because the SH_FILES and YM_FILES don't seem to get populated. They get populated and it work locally on my machine, with `$ bash -c 'echo $BASH_VERSION' 5.0.3(1)-release`. The travis one is 4.3.48. Maybe the `shopts` that I'm using are not present.

tests/Makefile
```diff
@@ -8,4 +8,9 @@ tests-catapult-integration:
...
 	./tests_catapult.sh
 
-all: tests-catapult tests-catapult-integration
\ No newline at end of file
+.PHONY: lint
+lint:
+	./lint.sh
+
+.PHONY: all
+all: lint tests-catapult tests-catapult-integration
```
mudler: Leave it off as default until we figure out why fails on travis, so we can merge anyway :)

viccuad:
Fixed the linting, now it's super pretty on travis. See the errors on `Fix regexp for buildir on tests/lint.sh` commit.

viccuad:
Works on travis, with `grep -v '^build'` (as it should). But in my machine I need `grep -v 'build*'`, which is wrong, as grep patterns are regexps and shouldn't work with blobs.

mudler:
Awesome! :stars:

PR #74, opened 2019-11-07
Cleanly remove data from cap-values

HartS:
- Deleting keys outright rather than setting them to the string "null"
restores the cap-values state to a pre-install state
- In the event that clean steps are run twice, attempting to use patch
to set a key to its current value will result in a failure, as the
patch command is surpsisingly *not* idempotent

### DIFF
--- a/modules/stratos/clean.sh
+++ b/modules/stratos/clean.sh
@@ -13,6 +13,9 @@ if kubectl get namespaces 2>/dev/null | grep -qi stratos ; then
 fi
 
 # delete STRATOS_CHART on cap-values configmap
+if [[ -n "$(kubectl get -n kube-system configmap cap-values | jq -r '.data["stratos-chart"]')" ]]; then
+    kubectl patch -n kube-system configmap cap-values --type json -p '[{"op": "remove", "path": "/data/stratos-chart"}]'
+fi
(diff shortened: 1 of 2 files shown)


modules/stratos/clean.sh
```diff
@@ -13,6 +13,9 @@ if kubectl get namespaces 2>/dev/null | grep -qi stratos ; then
 fi
 
 # delete STRATOS_CHART on cap-values configmap
+if [[ -n "$(kubectl get -n kube-system configmap cap-values | jq -r '.data["stratos-chart"]')" ]]; then
+    kubectl patch -n kube-system configmap cap-values --type json -p '[{"op": "remove", "path": "/data/stratos-chart"}]'
+fi
 kubectl patch -n kube-system configmap cap-values -p $'data:\n stratos-chart: "null"'
```
mudler: I think this is just a leftover, right?

viccuad:
Many thanks for making this PR! your approach is way cleaner :).

modules/stratos/clean.sh
viccuad: I concur here, the `stratos-chart: "null"` seems like a leftover.

HartS: Yep, in additional to that I forgot to specify `-o json` or the kubectl output. New commit addresses these issues

mudler:
Thank you ! Looks good here :)

PR #75, opened 2019-11-07
Add STORAGECLASS option on scf-gen-config target

viccuad:
Allows to select which storageclass are we going to use in scf-config-values.yaml


mudler:
That's really nice :dancer: will give a lot more flexibility depending on the backend!

*with the tinfoil hat on*: care to add a unit test ? :smiley:

viccuad:
Woops, thanks for the reminder, added! :)

PR #80, opened 2019-11-22
Use the current quarterly update JeOS image

dirkmueller:
This provides the latest SLE15 SP1 build for CaaSPv4 deployment.


dirkmueller:
(wip, trying to test..)

dirkmueller:
seems to work.

viccuad:
Many thanks! Deployed it also here, seems to work.
We definitely need a better way of having CaaSP systems in sync.

PR #81, opened 2019-12-02
Allow multi-master deployments

dirkmueller:
Previously the deployment failed with a weird docker error message
because it doesn't like newlines in environment variables. We can
tell jq to return the values as tab separated strings, which
is acceptable for docker.


viccuad:
Tested by deploying monomaster and multimaster deployments.

many thanks for the PR!

PR #83, opened 2019-12-02
Raise master nodes to 3 by default

dirkmueller:
We want to deploy multi-master on ECP to have a more realistic test
environment.


viccuad:
Many thanks for the PR!

PR #85, opened 2019-12-03
Default to SIZING=2 when HA=true

viccuad:
This fixes HA by providing at least 2 uaa instances when on HA.

From @dirkmueller's deployment: `ENABLE_EIRINI=false HA=true make` fails with
```
configmap/cap-values patched
Error: render error in "uaa/templates/uaa.yaml": template: uaa/templates/uaa.yaml:148:19: executing "uaa/templates/uaa.yaml" at <fail "uaa must have at least 2 instances for HA">: error calling fail: uaa must have at least 2 instances for HA
```

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export DEBUG_MODE="${DEBUG_MODE:-false}"
-source $ROOT_DIR/include/func.sh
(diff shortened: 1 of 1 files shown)


viccuad:
@mudler do we know if this will affect negatively the pipeline deployments?

mudler:
Looks good! ty

PR #86, opened 2019-12-04
Switch to kind 0.6.1

thardeck:


modules/scf/install.sh
```diff
@@ -17,6 +17,13 @@ if [[ $ENABLE_EIRINI == true ]] ; then
              --set args[0]="--kubelet-preferred-address-types=InternalIP" \
              --set args[1]="--kubelet-insecure-tls" || true
     fi
+
+    RUNS=0
+    until kubectl get --raw "/apis/metrics.k8s.io/v1beta1/nodes" || [ $RUNS -eq 24 ]; do
```
viccuad: Would it be possible to use something like `include/wait_ns.sh` but for pods, or add it to include?

backend/kind/kubeconfig.sh
```diff
@@ -4,6 +4,9 @@ set -ex
 . ../../include/common.sh
 . .envrc
 
-cp $(kind get kubeconfig-path --name="$CLUSTER_NAME") kubeconfig
+# Kind above version 0.6.0 generates the config itself
+if [ ! -f kubeconfig ]; then
+  cp $(kind get kubeconfig-path --name="$CLUSTER_NAME") kubeconfig
```
aojea: We are deprecating this option, check the release notes for the alternative, it will be removed soon https://github.com/kubernetes-sigs/kind/releases/tag/v0.6.0

include/common.sh
```diff
@@ -47,7 +47,7 @@ debug_mode
 export MAGICDNS="${MAGICDNS:-nip.io}"
 export ENABLE_EIRINI="${ENABLE_EIRINI:-true}"
 export EMBEDDED_UAA="${EMBEDDED_UAA:-false}"
-export KIND_APIVERSION="${KIND_APIVERSION:-kind.sigs.k8s.io/v1alpha2}"
+export KIND_APIVERSION="${KIND_APIVERSION:-kind.sigs.k8s.io/v1alpha3}"
```
aojea: is v1alpha4 since 0.6.0

thardeck: 3 seems to be backward compatible so it is easier to downgrade Kind to 0.2.1 for example by the environment variable. Is there an advantage to switch this to v4 by default?

mudler: We wanted to keep compatibility with 0.2.1 as well with this change, as some other automation scripts are depending on that version (still)

aojea: we should keep compatibility with the APIs.
The new API introduced some minor changes, moe the API namespace *.k8s.io  into x-k8s.io and the ability to patch kubeadm per node and containerd.

aojea: seems that you don't need it, was mostly a nit :sweat_smile:

PR #94, opened 2019-12-13
Add make scf-install as public target

viccuad:
No need to do `make private modules/scf install`, we can do `make scf-install`.

I'm a convert.


viccuad:
This could simplify calls in the pipelines, but of course it isn't needed.

PR #95, opened 2019-12-17
Fix and default to CAASP_VER=update in BACKEND=caasp4os

viccuad:
Default to CaaSP with update repositories when deploying it (CAASP_VER=update).

This PR build the skuba docker image into a sle15sp1, with all needed repos, product, and packages installed.
Patches the upstream CaaSP openstack terraform files to install the CaaSP product, update, and set swapaccount, and reboot, on all the nodes (see cloud-init/common.tpl and cloud-init/register-ibs.tpl).

### DIFF
--- a/backend/caasp4os/defaults.sh
+++ b/backend/caasp4os/defaults.sh
@@ -3,5 +3,5 @@
 # Caasp4os options
 ##################
 
-CAASP_VER=${CAASP_VER:-"product"}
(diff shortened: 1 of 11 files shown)


viccuad:
If testing this PR, please remember to delete your local existent `skuba/product` and `skuba/update` docker images. (`docker images` to get the sha, and then `docker rmi`).

PR #97, opened 2020-01-10
Let chart templates dictate count when HA is set

HartS:
Trying to set these manually in gen-config causes conflicts in HA,
where some roles have their count set to 1 when they should be
scaled, or 2 when the minimum required by the chart is 3


viccuad:
Many thanks for this PR :). Merging.

PR #98, opened 2020-01-10
Download binary deps as part of modules/common/deps.sh

### LINKED ISSUE #44: Add make target for downloading kubectl, helm, cfcli per build
Different k8s backends need different versions of helm, kubectl, etc: aws expects a specific version of kubectl depending on the k8s version you deploy, CaaSP should be pinned too, etc.

Add default HELM_VER, KUBECTL_VER, CFCLI_VER per backend, and save the version used for the whole life of the deployment in something like https://github.com/SUSE/catapult/issues/30.

As part of the `make buildir` or something similar, populate `build/bin/` with them.

By default, prefer the downloaded binaries. Add option to disable downloaded binaries or prefer local binaries by changing the concat order when creating `PATH` in `build/.envrc`.

viccuad:
Set default versions of `KUBECTL_VERSION`, `HELM_VERSION`, `CFCLI_VERSION`, and allow backends to also define a version that has more priority. This allows for example to pin the correct kubectl version depending on the k8s version targeted (see kind). 

Download `kubectl`, `helm`, `cf` by default. Add `DOWNLOAD_BINS` env var (default true) to possibly skip downloads.

Closes #44


viccuad:
This unblocks pinning caasp4os helm version to one matching `registry.suse.com/caasp/v4/helm-tiller`, which is the supported one.

Also helps first-time usage.

When merged, an update of https://github.com/SUSE/catapult/wiki/Local-requirements is needed.

viccuad:
Closes #44

Makefile
```diff
@@ -3,7 +3,7 @@ ifneq ($(DEBUG_MODE),true)
 endif
 
 # NOTE: BACKEND is dup in include/common.sh to allow BACKEND override when loading from json config files
-BACKEND?=kind
+export BACKEND?=kind
```
mudler: I cannot remember correctly now - have to check that out. But IIRC we had to remove the export here for the JSON config file support.

mudler: So it looks like: https://github.com/SUSE/catapult/commit/3696b195b1881d2c1670a360e445fba5674a345b . The export is happening in common/ so the JSON parameters can work. Why adding it back here?

viccuad: I exported BACKEND globally so it can be read from modules/common/deps.sh: https://github.com/SUSE/catapult/pull/98/files#diff-8ef793e96cb603b21d1eb01d85cab069R3-R5

We need BACKEND to do line 3:

```
line3   . "$ROOT_DIR/backend/$BACKEND/defaults.sh"
line4   . ./defaults.sh                             
line5   . ../../include/common.sh
```

viccuad:
Tested the changes succesfully with:
```
$ cat cap152.json
{
    "SCF_CHART": "https://s3.amazonaws.com/XXX/XXX/scf-sle-2.20.0-rc1%2Bcf12.17.0.0.g492d9cc2.zip",
    "ENABLE_EIRINI": "true",
    "DOCKER_ORG": "cap-XXX"
}
$ BACKEND=caasp4os CONFIG=$PWD/cap152.json make scf
```

Also, the json unittests pass, which is good.

mudler:
> Tested the changes succesfully with:
> 
> ```
> $ cat cap152.json
> {
>     "SCF_CHART": "https://s3.amazonaws.com/XXX/XXX/scf-sle-2.20.0-rc1%2Bcf12.17.0.0.g492d9cc2.zip",
>     "ENABLE_EIRINI": "true",
>     "DOCKER_ORG": "cap-XXX"
> }
> $ BACKEND=caasp4os CONFIG=$PWD/cap152.json make scf
> ```
> 
> Also, the json unittests pass, which is good.

Thanks for checking that out!

viccuad:
Many thanks for reviewing and providing pointers!

Updated https://github.com/SUSE/catapult/wiki/Local-requirements after the merge.

PR #100, opened 2020-01-22
Do not setup GARDEN_ROOTFS_DRIVER, just provide sane defaults

### LINKED ISSUE #99: Diego deployments defaults with btrfs
```
$> kubectl logs -f diego-cell-0 -n scf -c diego-cell
[...]
+ chmod 600 /var/vcap/data/grootfs/store/privileged/.backing-store
+ truncate -s 169034383360 /var/vcap/data/grootfs/store/privileged/.backing-store
+ mkfs.btrfs -f /var/vcap/data/grootfs/store/privileged/.backing-store
btrfs-progs v4.4.1
See http://btrfs.wiki.kernel.org for more information.

Label:              (null)
UUID:               af1eb860-6077-4736-9602-a463a29a20a2
Node size:          16384
Sector size:        4096
Filesystem size:    157.42GiB
+ local logfile=/var/log/create_and_mount_filesystem.log
+ mount -t btrfs -o loop,user_subvol_rm_allowed /var/vcap/data/grootfs/store/privileged/.backing-store /var/vcap/data/grootfs/store/privileged
Block group profiles:
  Data:             single            8.00MiB
  Metadata:         DUP               1.01GiB
  System:           DUP              12.00MiB
SSD detected:       no
Incompat features:  extref, skinny-metadata
Number of devices:  1
Devices:
   ID        SIZE  PATH
    1   157.42GiB  /var/vcap/data/grootfs/store/privileged/.backing-store

+ set +x
mount: /var/vcap/data/grootfs/store/privileged/.backing-store: failed to setup loop device: No such file or directory
Waiting for the loop kernel module to be loaded... done; bailing out for container restart.
```

https://github.com/SUSE/catapult/blob/master/modules/scf/defaults.sh#L14 invalidates https://github.com/SUSE/catapult/blob/master/modules/scf/gen_config.sh#L14 which at end draws a btrfs setup on diego which fails

mudler:
Eirini doesn't care about the Garden setting, prefer overlay-xfs for
default as it is more compatible.

Fixes #99

@viccuad not sure if this breaks something for caasp if it relies on `GARDEN_ROOTFS_DRIVER=btrfs` by defaault


mudler:
@viccuad probably https://github.com/SUSE/catapult/pull/100/commits/dbb9c9f929007ca0cd01c4e4d933671769382f32 should apply to other cases too, otherwise whatever we generated during the steps get ignored and only modules default gets applied

viccuad:
Thanks for looking into this, and sorry for needing some time before answering!

In theory, backend options do override *some* modules.
Example, for the global default options and modules/common:

`modules/common/defaults.sh` contains:
```
# Default versions in case the are undefined in the backend:
HELM_VERSION="${HELM_VERSION:-v2.16.1}"
KUBECTL_VERSION="${KUBECTL_VERSION:-v1.17.0}"
BAZEL_VERSION="${BAZEL_VERSION:-1.1.0}"
```
Yet, if any of those vars are defined on the backend, they will take precedence. This is achieved by the first line of the code on `modules/common/deps.sh`:
```
. "$ROOT_DIR/backend/$BACKEND/defaults.sh"
. ./defaults.sh
. ../../include/common.sh
. .envrc
```

This is cumbersome, but at least people reading `modules/common/deps.sh` and that snippet get an idea of the loading preference. Yet it doesn't use a compiled `<buildir>/defaults.sh`.

I would like to have a simpler approach, and maybe yours from this PR is it. I would love to save only changed options passed by env var on a different, `<buildir>/custom_options.sh`, that makes reviewing clusters way easier.

Can you think of an easier way to achieve this?

PR #105, opened 2020-02-03
Add yaml-patch as dependency

viccuad:
Add yaml-patch (https://github.com/krishicks/yaml-patch) as catapult dependency,
and use it for patching stratos and metrics configs, see `{stratos,metrics}/gen_config.sh`.

Add global option `DOWNLOAD_CATAPULT_DEPS`, to skip download of `bazel`, `yaml-patch` and the like, as ideally they should come from the system/docker image, to not download hundreds of MBs every time one creates a cluster.


viccuad:
This PR will need an update of https://github.com/SUSE/catapult/wiki/Local-requirements.

With DOWNLOAD_CATAPULT_DEPS we could maybe automate listing requirements anyways.

viccuad:
Rebased into master, so it can be used to test stratos comfortably.

PR #107, opened 2020-02-21
Add required config options to make stratos work on kubecf

greygoo:
- For Stratos we need to have a registry section and an external ip in the config.yaml in order to have `make stratos-gen-config` work  again and get a stratos deployment with an external ip to connect to.
- When enabling ingress for Stratos, a ingress hostname is required. In 1.5 this was retrieved from the DOMAIN value, but that does not exist for kubecf anymore it seems, thus I moved it into the console/service block.
- The storage class entry was removed as it seems to have no function for kubecf and does break the Stratos deployment as Stratos can't handle a null value here.

Alternatively it would also be possible to add the whole registry block via the stratos gen-config script and not have it created in the scf gen_config.sh The question here would be if it might also be beneficial for kubecf in catapult to have it in or not.

### DIFF
--- a/modules/scf/gen_config.sh
+++ b/modules/scf/gen_config.sh
@@ -54,9 +54,14 @@ features:
     enabled: ${ENABLE_EIRINI}
 
 kube:
-  storage_class: ~
   service_cluster_ip_range: 0.0.0.0/0
   pod_cluster_ip_range: 0.0.0.0/0
+
+  registry:
+    hostname: "${DOCKER_REGISTRY}"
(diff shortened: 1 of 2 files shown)


viccuad:
Thanks Chris for fixing this up! :)

Would you like to add [unit and integration tests](https://github.com/SUSE/catapult/tree/master/tests) for stratos? (that would make it easier to refactor, but more importantly, shifts the work of fixing stratos to the person that breaks it in the future : D )

PR #110, opened 2020-02-28
Make sure eks is using the correct kubectl version

greygoo:


viccuad:
Many thanks! this solves pesky errors!

PR #112, opened 2020-02-28
Fix extracting kubecf-bundle.tgz

greygoo:
I'm not sure this breaks something as I'm not fully sure if we still would have non bundled charts for kubecf.

### DIFF
--- a/modules/scf/chart.sh
+++ b/modules/scf/chart.sh
@@ -81,8 +81,10 @@ else
 fi
 
 if  [ "${SCF_OPERATOR}" == "true" ]; then
-    tar xvzf cf-operator*.tgz
-    tar xvzf kubecf*.tgz
+    if echo "$SCF_CHART" | grep -q "bundle"; then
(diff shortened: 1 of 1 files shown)


viccuad:
This PR is needed for consuming s3.kubecf-bundle correctly when `make scf-chart`

modules/scf/chart.sh
```diff
@@ -81,6 +81,8 @@ else
 fi
 
 if  [ "${SCF_OPERATOR}" == "true" ]; then
```
viccuad: I think this makes it work when consuming from s3.kubecf-bundle, but I think it will fail when consuming from s3.kubecf, as it doesn't have those files. Maybe an "if file exists, untar?"

greygoo: yes, what would be the better approach: 
- check for the existence of packaged files in the chart tarball and only extract them if they exist.
- check for the filename of the downloaded chart and handle in dependently

viccuad:
Perfect! many thanks!

PR #115, opened 2020-03-05
Add support for helm3

mudler:


mudler:
@viccuad as we are not sure if this break caasp4os workflow, we created a tag (helm2) in Catapult so it can be pinned in the pipelines. We tried to integrate the changes there too, but there might be leftovers

include/func.sh
```diff
@@ -236,8 +236,6 @@ function helm_install {
     local release_name=$1;shift
     local chart=$1;shift
     if [[ "$HELM_VERSION" == v3* ]]; then
-        info "Helm 3 version detected"
```
viccuad: :laughing: I see what you did there

PR #109, opened 2020-02-25
Add make kubeconfig to all backends, fix backends for product-release pipeline

viccuad:
Adds `BACKEND=<foo> make kubeconfig` for all backends.
Fixes some issues with CaaSP4 wrt to Helm3, externalIPs, etc.
Saves config creds of EKS, GKE on buildfolder/* (needed for pipelines).
Reevaluates GKE kubeconfigs to remove hardcoded paths of gcloud binaries.
Manually delete clusterroles, clusterrolebindings as helm3 doesn't seem to do that anymore.

PR #118, opened 2020-03-10
Fix quoting issue w external IPs in gen_config

HartS:
quoting the list of node IPs prevents them from separating into bash array elements

### DIFF
--- a/modules/scf/gen_config.sh
+++ b/modules/scf/gen_config.sh
@@ -20,7 +20,7 @@ fi
 
 domain=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["domain"]')
 public_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["public-ip"]')
-aux_external_ips=("$(kubectl get nodes -o json | jq -r '.items[].status.addresses[] | select(.type == "InternalIP").address')")
+aux_external_ips=($(kubectl get nodes -o json | jq -r '.items[].status.addresses[] | select(.type == "InternalIP").address'))
 external_ips+="\"$public_ip\""
 for (( i=0; i < ${#aux_external_ips[@]}; i++ )); do
 external_ips+=", \"${aux_external_ips[$i]}\""


mudler:
travis seems to complain for the opposite reason :roll_eyes:  I'm missing a bit of context, what's the real issue here?

HartS:
@mudler let's just take a look at what's happening here:

```
aux_external_ips=("$(kubectl get nodes -o json | jq -r '.items[].status.addresses[] | select(.type == "InternalIP").address')")
```

may result in something like the equivalent of:
```
aux_external_ips=("0.0.0.1
0.0.0.2
0.0.0.3")
```

The parentheses create a bash array! If all the IPs are quoted, this array then becomes an array with one element (a string with line breaks)

Later, this array is iterated over, but since there's only one element, this iteration doesn't work correctly

```
for (( i=0; i < ${#aux_external_ips[@]}; i++ )); do
external_ips+=", \"${aux_external_ips[$i]}\""
done
```

This won't cause problems with a single-node kubernetes, but it breaks `make scf-install` with multi-node kubernetes such as when `BACKEND=caasp4os`

mudler:
> @mudler let's just take a look at what's happening here:
> 
> ```
> aux_external_ips=("$(kubectl get nodes -o json | jq -r '.items[].status.addresses[] | select(.type == "InternalIP").address')")
> ```
> 
> may result in something like the equivalent of:
> 
> ```
> aux_external_ips=("0.0.0.1
> 0.0.0.2
> 0.0.0.3")
> ```
> 
> The parentheses create a bash array! If all the IPs are quoted, this array then becomes an array with one element (a string with line breaks)
> 
> Later, this array is iterated over, but since there's only one element, this iteration doesn't work correctly
> 
> ```
> for (( i=0; i < ${#aux_external_ips[@]}; i++ )); do
> external_ips+=", \"${aux_external_ips[$i]}\""
> done
> ```
> 
> This won't cause problems with a single-node kubernetes, but it breaks `make scf-install` with multi-node kubernetes such as when `BACKEND=caasp4os`

ouch, I see now, linting made us do unhappy choices here, we have to disable linting on this line. Thanks :100: for pointing that out and finding the culprit!

viccuad:
Argh! I fixed this on https://github.com/SUSE/catapult/commit/4aac0bc12c3e831776ae4373460d9610a484ed11

Sadly, because of the helm3 changes, the unit tests were not passing and I didn't not open a PR..

viccuad:
To be fair, just removing the quotes doesn't solve the issue (see [SC2207](https://github.com/koalaman/shellcheck/wiki/SC2207)), but it does get the desired behaviour. I may add my fix when I open a PR from the post-publish-spike branch. Not having the fix already on master falls on me, though. Sorry.

mudler:
Shall we close this in favor of #109 ?

viccuad:
> Shall we close this in favor of #109 ?

Sadly, yes. But thanks @HartS , and sorry for the double work :disappointed:

PR #119, opened 2020-03-11
Remove unneeded logic to obtain the skuba version, use date(1)

viccuad:
We used to start a transient sle15 image, install skuba, and obtain the version.
Something has changed and it fails.

We don't care so much in obtaining the version number to just pass it as a
build-arg VERSION when creating the skuba/foo image, because it moves slowly,
and we normally only need a skuba/update image, not a skuba/product, or a
skuba/devel image.

Let's just remove it.

Thanks to a triplet programming session, between Hart, Prabal and Victor.

### DIFF
--- a/backend/caasp4os/docker/skuba/build.sh
+++ b/backend/caasp4os/docker/skuba/build.sh
@@ -2,37 +2,12 @@
 
 REPO_ENV="$1"
 REPO="$2"
-VERSION=
+IMAGE_NAME="skuba/$REPO_ENV"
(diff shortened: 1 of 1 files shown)


viccuad:
This was tested yesterday as we talked and found the fix, by cleaning my local sle 15 image and deploying caasp.

@prabalsharma @HartS merging this without your reviews, people are having trouble trying to deploy CaaSP4 from scratch, and this fixes it.

PR #133, opened 2020-03-13
Consume kubecf released charts from github when SCF_CHART not set

viccuad:
Now that we have a kubecf release (yaay!), do as we did with scf.

Consume specific charts:
  `SCF_OPERATOR=true SCF_CHART=<url or abs path> make scf`

By default, if no SCF_CHART is passed, consume latest released public chart from
github:
  `SCF_OPERATOR=true make scf`


viccuad:
Tested that it works on kind with `make k8s; SCF_OPERATOR=true make scf`

viccuad:
merging, as some people are finding it cumbersome to need to provide the chart all the time

Issue #128, opened 2020-03-13
Some kubecf resources not cleaned up by `make scf-clean`


HartS:
Testing kubecf deployments via catapult, when I clean a kubecf cluster with `make scf-clean`, and attempt to redeploy with `make scf`, I'll encounter this error:

```
Installing CFO from: https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-3.2.1%2B0.ga32a3f79.tgz
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: , name: cf-operator-quarks-job, existing_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole, new_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole
Makefile:17: recipe for target 'install' failed
```

I found I have to delete the CR and CRBs manually with the following commands before I can deploy again:

```
kubectl delete clusterrolebinding cf-operator cf-operator-quarks-job
kubectl delete clusterrole cf-operator-quarks-job cf-operator
```

And in a previous reinstall, I also had to run:
```
kubectl delete psp susecf-scf-default
```
(though I didn't see this psp lingering after the last clean)

viccuad:
Thanks for reporting this!

I can reproduce this, and the cap-release pipeline too.

This was not picked up by the kubecf pipeline because they spawn kind clusters, use them, and then get rid of them.

Apart from those changes, once one configures kubecf to deploy with the qjobs for testsuites, with `SCF_TESTGROUP=true` (see modules/scf/gen-config.sh), it is needed to remove these too, and maybe more for all testsuites:

```
kubectl delete clusterrole scf-susecf-scf-test-cluster-role || true                                                                                                                                                  
kubectl delete clusterrolebinding \                                                                                                                                                                                  
        scf-susecf-scf-tests-brain-test-cluster-role-cluster-binding || true                                                                                                                                         
kubectl delete clusterrolebinding \                                                                                                                                                                                  
        scf-susecf-scf-tests-sits-test-cluster-role-cluster-binding || true 
```

I have removed them by hand and redeployed, and weirdly kubecf complains on deployment that they exist. It is like the kubecf helm chart adds them twice, and then complains it exists.


All these deletions could be added to `modules/scf/clean.sh`, do you feel like tackling it?

viccuad:
[...]

PR #135, opened 2020-03-16
Add helm3 support everywhere

viccuad:
Add and or consume helm_{init_client, ls, install, upgrade} functions as needed.

Implement changes from:                                                                                                                                                                                                                                                                
* https://helm.sh/docs/topics/v2_v3_migration/                                                                                                                                                                                                                                             
* https://helm.sh/docs/faq/

### DIFF
--- a/include/buildir.sh
+++ b/include/buildir.sh
@@ -24,7 +24,13 @@ HEREDOC
 
 cat <<HEREDOC_APPEND >> .envrc
 export KUBECONFIG="$(pwd)"/kubeconfig
-export HELM_HOME="$(pwd)"/.helm
+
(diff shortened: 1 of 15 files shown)


viccuad:
Tested on kind and CaaSP deployments several times, for kubecf, stratos, metrics, scf.

I had a bit of faith that it would solve https://github.com/SUSE/catapult/issues/128, but no.
It seems that one needs to wait for a bit after a helm_delete when helm3, as helm3 doesn't do a  --wait anymore. I don't consider it a regression, since it is only on the new functionality of helm3.

mudler:
Looks good here! ty! :100:

viccuad:
many thanks for the review!

PR #141, opened 2020-03-17
Mark scf-deploy target deprecated, don't print it on make help-all

viccuad:


mudler:
I think the make target is still handy. Why removing the target that makes the scf deployment as a separate step?
  
I thought we were about to remove all the cleanup steps **before** it, by reducing the scope and just do the "deploy"

viccuad:
The issue is that `scf-deploy` cleans the cluster. For `scf-deploy` to not remove the cluster, it should not call `clean` or `k8s`. Which then it's the same behaviour as `make scf`.

If so, why keep it around?

PR #142, opened 2020-03-20
Use kubectl delete with --ignore-not-found

HartS:


viccuad:
many thanks!
This makes the `if` checks redundant, removing them simplifies it and makes it faster. Do you feel like tackling that in this PR? if not, just feel free to merge this :)

PR #143, opened 2020-03-20
Remove deprecated SCF tests

HartS:
These tests aren't referenced from Makefile, and are no longer relevant with
kubecf.


mudler:
Those tests are for catapult itself. You can run them with `make catapult-test`

viccuad:
The unit and integration tests contain things that are not only scf, for example the buildir creation, downloading of needed binaries, backend unit tests, stratos, and stratos-metrics.

I would be glad to remove them, but I don't think we should remove unit and integration tests unless we remove the code that they test, and in the case of scf, we should have an alternative in place before.

PR #148, opened 2020-03-30
Wait for cf-operator resources to be available before start KubeCF deployment

mudler:
This should help with https://github.com/cloudfoundry-incubator/kubecf/issues/558 . It's hard to reproduce, and I couldn't locally, but waiting for the operator crd to be in place seems safer than start deploying right away.

### DIFF
--- a/include/func.sh
+++ b/include/func.sh
@@ -281,3 +281,8 @@ function helm_ls {
         helm ls
     fi
 }
+
+function wait_for {
+    info "Waiting for $1"
+    n=0; until ((n >= 60)); do eval "$1" && break; n=$((n + 1)); sleep 1; done; ((n < 60))
+}
--- a/modules/scf/install.sh
+++ b/modules/scf/install.sh
@@ -64,7 +64,15 @@ elif [ "${SCF_OPERATOR}" == "true" ]; then
     --set "global.operator.watchNamespace=scf"
 
     wait_ns cf-operator
-    sleep 10
+
+    info "Wait for cf-operator to be ready"
+    wait_for "kubectl get endpoints -n cf-operator cf-operator-webhook -o name"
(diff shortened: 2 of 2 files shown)


modules/scf/install.sh
```diff
@@ -64,7 +64,15 @@ elif [ "${SCF_OPERATOR}" == "true" ]; then
...
     wait_ns cf-operator
-    sleep 10
+
+    info "Wait for cf-operator to be ready"
+    wait_for "kubectl get endpoints -n cf-operator cf-operator-webhook -o name"
+    wait_for "kubectl get crd quarksstatefulsets.quarks.cloudfoundry.org -o name"
+    wait_for "kubectl get crd quarkssecrets.quarks.cloudfoundry.org -o name"
+    wait_for "kubectl get crd quarksjobs.quarks.cloudfoundry.org -o name"
+    wait_for "kubectl get crd boshdeployments.quarks.cloudfoundry.org -o name"
+    sleep 10 # Give extra time for operator to avoid flakyness
```
jimmykarily: Magic number are bad but if we can't avoid them maybe we should mention in the comment what exactly are we waiting for. This way people will be able to remove the "sleep" in the future by adding more checks like the above.

mudler: It is there just to avoid random flakiness from the operator (as the comment already suggests) as it might be not ready yet. It's really hard to reproduce locally, the new checks might make the sleep here useless but I would rather avoid removing it right now

include/func.sh
```diff
@@ -281,3 +281,8 @@ function helm_ls {
         helm ls
     fi
 }
+
+function wait_for {
+    info "Waiting for $1"
+    n=0; until ((n >= 60)); do eval "$1" && break; n=$((n + 1)); sleep 1; done; ((n < 60))
+}
```
bikramnehra: How about a newline here?

PR #150, opened 2020-04-01
Change cf-operator --provider flag to --operator-webhook-use-service-reference

viccuad:
See:
https://github.com/cloudfoundry-incubator/cf-operator/pull/616

Signed-off-by: Christian Hueller <chuller@suse.com>


mudler:
Is there some way to make this backward compatible too? I already see a lot of sad pandas after this office

viccuad:
The cf-operator flag change is from October 2019, it is included on cf-operator since 0.4.2 and up to current 4.0.0 (see tags on https://github.com/cloudfoundry-incubator/cf-operator/pull/616/commits/ce26da036d45e676c31fbaef065fe82fd93a8a5b) so I don't see much need to make this backwards compatible.

PR #151, opened 2020-04-02
kubecf-test: compatibility with cf-operator 4

mook-as:
The new cf-operator no longer prefixes resources (statefulsets, pods, etc.) with the deployment name.  Update the script to handle that case (while also supporting the older version).

Also, refactor the script a bit to be (hopefully) more readable, or at least have shorter lines.

### DIFF
--- a/modules/tests/kubecf-test.sh
+++ b/modules/tests/kubecf-test.sh
@@ -1,82 +1,95 @@
 #!/bin/bash
 
+# This script runs a test suite on kubecf
+
 . ./defaults.sh
 . ../../include/common.sh
 . .envrc
 
-smoke_tests_pod_name() {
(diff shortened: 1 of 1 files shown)


mudler:
P.S. I've just checked and you seems to have permission to write in this repo. It's much better to use branches when opening PRs so you can also pin the catapult resource from the pipeline directly

mook-as:
The patching is (hopefully) fixed. Not sure how to test on concourse; while I updated the [branch](https://github.com/SUSE/catapult/commits/mook-as-operator-4-compat), I can't seem to pin in in the [resource](https://concourse.suse.dev/teams/main/pipelines/kubecf/resources/catapult) (which appears to just have the `master` commits?).  Tried to override it explicitly in the pipeline definition, but that just gets the step stuck on `waiting for a suitable set of input versions / catapult - no versions available`.  (I'm reverting that change now.)

PR #154, opened 2020-04-03
Remove test jobs if successful and save test logs

viccuad:
If test job successful, remove job (which deletes the pod). This is needed so following `helm upgrade` (with `make scf-upgrade`) on the pipelines work as they don't redeploy and rerun the tests.

Also, whatever the outcome of the test job, save its logs onto `buildfolder/artifacts/{date}_{podname}.log`, eg:
`buildfolder/artifacts/2020-04-03-11:30_susecf-scf-smoke-tests-9edf0682ed6cf030-qh7mk.log`

### DIFF
--- a/modules/tests/kubecf-test.sh
+++ b/modules/tests/kubecf-test.sh
@@ -96,10 +96,23 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
 
 # Wait for the container to terminate and then exit the script with the container's exit code.
 jsonpath='{.status.containerStatuses[?(@.name == "'"$container_name"'")].state.terminated.exitCode}'
-while true; do
-  exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
-  if [[ -n "${exit_code}" ]]; then
-      exit "${exit_code}"
-  fi
-  sleep 1
+exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
+while [[ -z "${exit_code}" ]]; do
+    exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
+    sleep 1
 done
+
+# save results of tests to file
+mkdir -p artifacts
+kubectl logs "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$container_name" > \
+        artifacts/"$(date +'%Y-%m-%d-%H:%M')"_"$( echo "${pod_name}" | cut -d '/' -f 2)".log
+
+if [ "${exit_code}" -ne 0 ]; then
+    err "${KUBECF_TEST_SUITE} failed"
+    exit "${exit_code}"
+else
+    # remove job, tests where successful
+    kubectl delete jobs -n "${KUBECF_NAMESPACE}"  --all --wait || true
+fi
+
+ok "${KUBECF_TEST_SUITE} passed"
--- a/modules/tests/kubecf-test.sh
+++ b/modules/tests/kubecf-test.sh
@@ -96,7 +96,7 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
 
 # Wait for the container to terminate and then exit the script with the container's exit code.
 jsonpath='{.status.containerStatuses[?(@.name == "'"$container_name"'")].state.terminated.exitCode}'
-exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
+exit_code=""
 while [[ -z "${exit_code}" ]]; do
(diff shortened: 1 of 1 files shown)


modules/tests/kubecf-test.sh
```diff
@@ -96,10 +96,23 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
...
 # Wait for the container to terminate and then exit the script with the container's exit code.
 jsonpath='{.status.containerStatuses[?(@.name == "'"$container_name"'")].state.terminated.exitCode}'
-while true; do
-  exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
-  if [[ -n "${exit_code}" ]]; then
-      exit "${exit_code}"
-  fi
-  sleep 1
+exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
+while [[ -z "${exit_code}" ]]; do
```
mudler, line 100: Wouldn't make sense to re-use the same approach used in SCF cats tests? https://github.com/SUSE/catapult/blob/master/modules/tests/kubecats.sh#L37

viccuad: doh, we didn't realize. thanks!

modules/tests/kubecf-test.sh
```diff
@@ -96,10 +96,23 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
...
+    exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
+    sleep 1
 done
```
mook-as, line 103: (In addition to re-using the functions that I didn't realize existed) we could also just force things to loop at least once to make the code simpler:
```suggestion
exit_code=""
while [[ -z "${exit_code}" ]]; do
    exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
    sleep 1
done
```

Or we if we really want to not sleep excessively, we can do something like the old code:
```suggestion
while true; do
  exit_code="$(kubectl get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
  if [[ -n "${exit_code}" ]]; then
    break
  fi
  sleep 1
done
```

modules/tests/kubecf-test.sh
```diff
@@ -96,10 +96,23 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
...
+
+# save results of tests to file
+mkdir -p artifacts
+kubectl logs "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$container_name" > \
+        artifacts/"$(date +'%Y-%m-%d-%H:%M')"_"$( echo "${pod_name}" | cut -d '/' -f 2)".log
```
mook-as, line 108: This line is a bit long; maybe use a temporary variable?
```suggestion
log="artifacts/$(date +'%Y-%m-%d-%H:%M')_${pod_name#*/}.log"
kubectl logs "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$container_name" > "${log}"
        
```

viccuad: way cleaner, and with var substitution, thanks.

modules/tests/kubecf-test.sh
viccuad: `wait_container_attached()` is not useful as we need to save the exit code of the container. Going with Mark's simplifying suggestion.

modules/tests/kubecf-test.sh
viccuad: Going with your first suggestion; we need to save the exit_code of the pod after the wait on the while anyways.

modules/tests/kubecf-test.sh
```diff
@@ -96,10 +96,24 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
...
+# save results of tests to file
+mkdir -p artifacts
+log="artifacts/$(date +'%Y-%m-%d-%H:%M')_${pod_name#*/}.log"
+kubectl logs "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$container_name" > "${log}"
+
+
+if [ "${exit_code}" -ne 0 ]; then
+    err "${KUBECF_TEST_SUITE} failed"
+    exit "${exit_code}"
+else
```
mook-as, line 114:
```suggestion
fi
```

There's no point in an `else` given that the other branch `exit`s.

modules/tests/kubecf-test.sh
```diff
@@ -96,10 +96,24 @@ kubectl logs -f "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container "$co
...
+    # remove job, tests where successful
```
mook-as, line 115:
```suggestion
    # remove job, tests were successful
```

modules/tests/kubecf-test.sh
viccuad: done, thanks!

mudler:
> Should still wait for @mudler to do a proper review though, I think he's more familiar with the code. I'm just treating this as generic shell script :)

That's perfectly fine :smiley: !

PR #155, opened 2020-04-06
Add catapult deps and backend deps to Dockerfile

viccuad:
Add to Dockerfile:
  - catapult deps: `yq`, `bazel`, `yaml-patch`. Add one version of `kubectl, cf-cli` each, for convenience
  - k8s provider deps: `aws-cli`, `aws iam`, `gcloud`, `az`, `terraform`

Clean and reduce catapult image

Don't dowload deps if they are already in `buildfolder/bin/*`

Use `DOWNLOAD_CATAPULT_DEPS` everywhere, which allows pipelines to skip downloads. Default to true so normal users will download deps.

PR #156, opened 2020-04-06
Cats internetless

jimmykarily:
This adds a target to run the cats internetless suite. It is needed for this: https://github.com/cloudfoundry-incubator/kubecf/issues/144

### DIFF
--- a/backend/kind/up.sh
+++ b/backend/kind/up.sh
@@ -19,8 +19,13 @@ nodes:
   - containerPath: /code
     hostPath: ${APPLICATION_PATH}
     # readOnly: true
+networking:
+  disableDefaultCNI: true
 EOF
 
 kind create cluster --config kind-config.yaml --name=${CLUSTER_NAME}
+
+# TODO: Run this somewhere to enable weave network plugin:
+kubectl apply -f https://github.com/weaveworks/weave/releases/download/v2.6.2/weave-daemonset-k8s-1.11.yaml
 
 ok "Cluster is up"
--- a/backend/kind/up.sh
+++ b/backend/kind/up.sh
@@ -25,7 +25,9 @@ EOF
 
 kind create cluster --config kind-config.yaml --name=${CLUSTER_NAME}
 
-# TODO: Run this somewhere to enable weave network plugin:
+# Since we use disableDefaultCNI above we need a network plugin
+# weave allows us to use NetworkPolicy to block internet access for
+# CATs internetless suite or other use.
 kubectl apply -f https://github.com/weaveworks/weave/releases/download/v2.6.2/weave-daemonset-k8s-1.11.yaml
 
 ok "Cluster is up"
--- a/modules/common/deps.sh
+++ b/modules/common/deps.sh
@@ -67,7 +67,7 @@ else
 
     yqpath=bin/yq
     if [ ! -e "$yqpath" ]; then
-        wget "https://github.com/mikefarah/yq/releases/download/2.4.0/yq_${YQ_OS_TYPE}" -O $yqpath
+        wget "https://github.com/mikefarah/yq/releases/download/3.2.1/yq_${YQ_OS_TYPE}" -O $yqpath 
         chmod +x $yqpath
     fi
 fi
--- a/modules/tests/kubecf-test.sh
+++ b/modules/tests/kubecf-test.sh
@@ -136,11 +136,12 @@ create_cats_internetless_secret() {
     kubectl get secret -n "${KUBECF_NAMESPACE}" "${cats_secret_name}" -o json  | \
     jq -r '.data."properties.yaml"' | base64 -d | \
     yq w - "instance_groups.(name==acceptance-tests).jobs.(name==acceptance-tests).properties.acceptance_tests.include" '=internetless' | \
+    yq w - "instance_groups.(name==acceptance-tests).jobs.(name==acceptance-tests).properties.acceptance_tests.credhub_mode" '"off"' | \
     base64 -w 0 )"
 
-    cats_updated="$(echo "$cats_secret" | yq w - 'data[properties.yaml]' $cats_updated_properties)"
-    cats_updated="$(echo "$cats_updated" | yq w - 'metadata.name' 'cats-internetless')"
-    kubectl apply -f <(echo "${cats_updated}") -n "${KUBECF_NAMESPACE}"
+  cats_updated="$(echo "$cats_secret" | yq w - 'data[properties.yaml]' $cats_updated_properties)"
+  cats_updated="$(echo "$cats_updated" | yq w - 'metadata.name' 'cats-internetless')"
+  kubectl apply -f <(echo "${cats_updated}") -n "${KUBECF_NAMESPACE}"
 }
 
 mount_cats_internetless_secret() {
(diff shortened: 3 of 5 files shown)


backend/kind/up.sh
```diff
@@ -19,8 +19,13 @@ nodes:
...
     hostPath: ${APPLICATION_PATH}
     # readOnly: true
+networking:
+  disableDefaultCNI: true
 EOF
 
 kind create cluster --config kind-config.yaml --name=${CLUSTER_NAME}
 
+# TODO: Run this somewhere to enable weave network plugin:
+kubectl apply -f https://github.com/weaveworks/weave/releases/download/v2.6.2/weave-daemonset-k8s-1.11.yaml
```
jimmykarily, line 29: I wonder if this is ok here. I will run with a fresh cluster to see if I end up with a working weave pod.

jimmykarily:
Output: 

```
Running Suite: CATS
===================
Random Seed: 1586176372
Will run 213 specs

Running in parallel across 2 nodes

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSPSSSSSSSSSSSSSSSSSSSSSSS
------------------------------
Running CATs with CF CLI version  cf version 6.46.1+4934877ec.2019-08-23


SSSSSSSSSS•
Ran 1 of 213 Specs in 88.962 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Flaked | 1 Pending | 211 Skipped


Ginkgo ran 1 suite in 1m29.679860433s
Test Suite Passed
Unbinding security group loopback-secgroup-1586176347 from defaults for staging as admin
OK



TIP: Changes will not apply to existing running applications until they are restarted.
Unbinding security group loopback-secgroup-1586176347 from defaults for running as admin
OK



TIP: Changes will not apply to existing running applications until they are restarted.
Deleting security group loopback-secgroup-1586176347 as admin
OK
Configure cluster network: Full access
networkpolicy.networking.k8s.io "cats-internetless" deleted
Mounting the original secret in acceptance tests qjob
quarksjob.quarks.cloudfoundry.org/susecf-scf-acceptance-tests patched
secret "cats-internetless" deleted
```

backend/kind/up.sh
jimmykarily: Works fine here. @mudler is there a better place to put this? Otherwise I will keep it here. This will be always applied no matter what you are planning to use your cluster for. Seems to work for my case here but I wonder if we want this in every use case.

modules/common/deps.sh
```diff
@@ -67,7 +67,7 @@ else
 
     yqpath=bin/yq
     if [ ! -e "$yqpath" ]; then
-        wget "https://github.com/mikefarah/yq/releases/download/2.4.0/yq_${YQ_OS_TYPE}" -O $yqpath
+        wget "https://github.com/mikefarah/yq/releases/download/3.2.1/yq_${YQ_OS_TYPE}" -O $yqpath
```
jimmykarily, line 70: I remember we needed some new latest feature of `yq` and that's why we bumped the version. I don't remember what and I can't see anything special being called in this PR, maybe we removed the special code in the end. Should I still keep this bump? Won't hurt I guess.

modules/tests/kubecf-test.sh
```diff
@@ -58,6 +58,118 @@ wait_for_tests_pod() {
...
+# Enables internet access again and mounte the original secret in the qjob
+# to enable other cats again.
+cleanup_cats_internetless() {
+  rv=$?
+  isolate_network 0
+  revert_patch='{ "spec": { "template": { "spec": { "template": { "spec": { "volumes": '${original_volumes}' } } } } } }'
+  echo "$(blue "Mounting the original secret in acceptance tests qjob")"
+  kubectl patch qjob "${KUBECF_DEPLOYMENT_NAME}"-acceptance-tests --namespace "${KUBECF_NAMESPACE}" --type merge --patch "${revert_patch}"
+  kubectl delete secret -n ${KUBECF_NAMESPACE} cats-internetless
+  trap "exit \$rv" EXIT
```
jimmykarily, line 170: This is to preserve the original exit code and not return whatever the cleanup function returns. I remember we did that some other place too but I don't remember if we used the same "trick". This one seems to work.

backend/kind/up.sh
mudler: I don't think we can do much more, otherwise we need (yet another) flag to opt-in for this feature explictly. If doesn't hurt as default, I would keep as it is (better opinionated than giving too many options).

The only thing is, we need also to make the same change at the ekcp backend in Catapult. Note we can post also a kindconfig (is not documented anywhere.. my bad) https://github.com/mudler/ekcp/blob/b585bb2a22393f9b870651da06428917fdd6cf7c/api.go#L36 , so it should work somehow like this:

`curl -d "name=test&kindconfig=$(base64 kindconfig.yaml)" -X POST http://127.0.0.1:8030/new`

modules/common/deps.sh
mudler: I'm fine with the bump ! :+1:

viccuad: the highest version available in Tumbleweed/Leap is 2.10; if we don't need it I wouldn't bump it.

If we do, we need to consume 3.2.1 binary in the Dockerfile, maybe at https://github.com/SUSE/catapult/pull/155.

modules/tests/kubecf-test.sh
```diff
@@ -58,6 +58,118 @@ wait_for_tests_pod() {
...
+      # For debugging, show what kube thinks of it.
+      kubectl describe networkpolicies \
+        --namespace "${KUBECF_NAMESPACE}" \
+        cats-internetless
+    else
+      # shellcheck disable=SC2005
+      echo "$(blue "Configure cluster network: Full access")"
+      # disable isolation
+      kubectl delete networkpolicies \
+        --namespace "${KUBECF_NAMESPACE}" \
```
viccuad, line 121: This will fail if network policy isn't present, can be made idempotent:

```suggestion
        --namespace "${KUBECF_NAMESPACE}" \
        --ignore-not-found \
```

modules/tests/kubecf-test.sh
```diff
@@ -58,6 +58,118 @@ wait_for_tests_pod() {
...
+    jq -r 'map((select(.name=="ig-resolved") | .secret.secretName) |= "cats-internetless")')
+
+  patch='{ "spec": { "template": { "spec": { "template": { "spec": { "volumes": '${updated_volumes}' } } } } } }'
+
+  kubectl patch qjob "${KUBECF_DEPLOYMENT_NAME}"-acceptance-tests \
+    --namespace "${KUBECF_NAMESPACE}" --type merge --patch "${patch}"
+}
+
+# Enables internet access again and mounte the original secret in the qjob
+# to enable other cats again.
```
viccuad, line 162:
```suggestion
# Re-enables internet access and mounts the original secret in the qjob
# to allow internet-full cats.
```

modules/common/deps.sh
jimmykarily: Turns out we need a newer version. With the old one we get this in the cats pod : 


```
│ + cf-operator util template-render                                                                                                                                                          │
│ 2020/04/07 12:03:43 failed to load BOSH deployment manifest /var/run/secrets/resolved-properties/acceptance-tests/properties.yaml: failed to unmarshal BOSH deployment manifest director_uu │
│ id: ""                                                                                                                                                                                      │
│ instance_groups:                                                                                                                                                                            │
│   (name==acceptance-tests):                                                                                                                                                                 │
│     jobs:                                                                                                                                                                                   │
│       (name==acceptance-tests):                                                                                                                                                             │
│         properties:                                                                                                                                                                         │
│           acceptance_tests:                                                                                                                                                                 │
│             include: =internetless                                                                                                                                                          │
│             credhub_mode: "off"                                                                                                                                                             │
│ : error unmarshaling JSON: while decoding JSON: json: cannot unmarshal object into Go struct field Manifest.instance_groups of type manifest.InstanceGroups    
```

jimmykarily: Seems to be coming from this which doesn't get interpreted by yq: https://github.com/SUSE/catapult/pull/156/files#diff-e363fc0b9655dc80f5abcfaafbe11d82R138

mudler:
For reference, to consume it in KubeCF CI we need to address https://github.com/SUSE/catapult/pull/156#discussion_r404076167 , but shouldn't prevent it from merging

mudler:
Created a new issue https://github.com/SUSE/catapult/issues/161 for the CI work

viccuad:
merged https://github.com/SUSE/catapult/pull/154, so this PR can be rebased on top of master. But to be honest, I see no conflict.

viccuad:
Also, how it is that only 1 test of 213 is run? does it make sense to consume it then?

jimmykarily:
There is only one test afaict in the "internetless suite": https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/internetless/app_external_traffic_denied.go#L60

jimmykarily:
Rebased and fixed all comments on this PR. I think we'd better wait until the ekcp issue is also resolved before we merge. Otherwise we won't be able to bump catapult in the pipelines.

mudler:
I think we can merge safely, the CI is using ekcp backend, not kind, so the weave change won't impact CI runs

viovanov:
@jimmykarily @mudler can you please explain exactly what we need in order to unblock https://github.com/cloudfoundry-incubator/kubecf/issues/144 ?

mudler:
If the outcome of  https://github.com/cloudfoundry-incubator/kubecf/issues/144 is to add a step in the pipeline, then needs https://github.com/SUSE/catapult/issues/161 to get solved. But I think it can be treated separately, as this PR doesn't touches the backend which is used by the CI.

This PR applies weave only to the kind backend, which means that the test could be run only locally (but not by Concourse).

PR #157, opened 2020-04-06
Use JeOS SLE15 SP1 QU2

dirkmueller:
This is a newer base image refreshed with maintenance
updates released up to end of january.


viccuad:
Many thanks! merging.

This will shorten the time to deploy caasp4 on openstack, but note that deployments are already up to date, as we default to update repos: https://github.com/SUSE/catapult/blob/e52e7c7f427003599539daeab08af62246a37407/backend/caasp4os/deploy.sh#L60

PR #160, opened 2020-04-07
Added support for parametrization of brain tests.

andreas-kupries:
Ref: https://jira.suse.com/browse/CAP-1353
See also https://github.com/SUSE/cap/pull/4 and https://github.com/SUSE/brain-tests-release/pull/12

Defaults for standard setup.
Changes inserted via kubecf property override mechanism.

Testing: Manual run of catapult `make all` using the configuration

```
export BRAIN_INCLUDE=010
export BRAIN_VERBOSE=true
export CLUSTER_NAME=brainy
export BACKEND=minikube
export ENABLE_EIRINI=false
export SCF_TESTGROUP=true
export KUBECF_TEST_SUITE=brain
export SCF_OPERATOR=true
```

While this did not fully deploy the kubecf (*) it came far enough that is was possible to check the user provided properties, i.e. use `kubectl get configmap -n scf -o yaml susecf-scf-user-provided-properties` to confirm that catapult generated the proper settings.

These settings had been separately tested against a non-catapult minikube as part of https://github.com/SUSE/brain-tests-release/pull/12 . The BTR PR is required to fix issues in the BTR itself able to mis-transmit certain include/exclude patterns.

(*) Stuck with diego-cell errors and credhub Init:7/13. IOW outside of the brain test config changes.


modules/scf/gen_config.sh
```diff
@@ -112,6 +112,13 @@ properties:
           slow_spec_threshold: 300
           nodes: 2
           flake_attempts: 4
+  brain-tests:
+    acceptance-tests-brain:
+      acceptance_tests_brain:
+        verbose: ${BRAIN_VERBOSE}
```
viccuad, line 118: Nitpicking: this vars aren't quoted because it's not needed, yet maybe it's a good idea to quote them

andreas-kupries:
@mudler Will do. Thanks for the reference.

modules/scf/gen_config.sh
andreas-kupries: Sure. Should not hurt.

andreas-kupries: Addressed with commit 78b25e6730a5ff609b295d14b48e5b8953fe9a05

PR #162, opened 2020-04-08
Revert  and fix "Revert "Add catapult deps and backend deps to Dockerfile""

viccuad:
Reverts SUSE/catapult#159.

Pins package to their repos correctly so the Dockerfile builds even if there's dependency changes.
Fixes typo on Dockerfile wrt --no-recommends.


viccuad:
Tested by building a Dockerfile locally, all fine.

viccuad:
Tested manually by running `make catapult-image`, docker image builds and contains everything correctly. Finally!

PR #163, opened 2020-04-08
Make sure br_netfilter is loaded so that kind works

dirkmueller:
Without that, there is this error:
  cat: can't open '/proc/sys/net/bridge/bridge-nf-call-iptables': No such file or directory


dirkmueller:
(untested)

PR #164, opened 2020-04-09
Consume MAGICDNS on m/stratos/gen-config.sh, don't hardcode nip.io

viccuad:


viccuad:
Better than this, maybe we should be consuming the domain of the cluster from the `cap-values` configmap. But this PR doesn't hurt for now.

mudler:
Superseded by #170 I guess?

viccuad:
Correct, sorry. Closing!

PR #169, opened 2020-04-13
enable gke cluster name, update to catapult docker image

prabalsharma:
catapult was only setting label for gke cluster. With this PR, catapult will set cluster name along with label for gke cluster. This is required for CAP CI work, where we want to control cluster name and label.

This PR also fixes path for gcloud cli and adds helm to catapult docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,13 +30,10 @@ RUN curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.
 RUN curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator && \
   chmod +x aws-iam-authenticator && mv aws-iam-authenticator /usr/local/bin/
 
-RUN curl -o google-cloud-sdk.tar.gz https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-264.0.0-linux-x86_64.tar.gz && \
-  tar -xvf google-cloud-sdk.tar.gz && \
-  rm google-cloud-sdk.tar.gz && \
-  pushd google-cloud-sdk || exit && \
-  bash ./install.sh -q && \
-  popd || exit && \
-  echo "source /google-cloud-sdk/path.bash.inc" >> ~/.bashrc
+RUN curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz \
+&& mkdir -p /usr/local/gcloud \
+&& tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
+&& /usr/local/gcloud/google-cloud-sdk/install.sh --quiet
 
 RUN curl -o kubectl-aws https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/kubectl && \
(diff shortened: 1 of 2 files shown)


backend/gke/deploy.sh
```diff
@@ -22,6 +22,7 @@ fi
 
 git clone https://github.com/SUSE/cap-terraform.git
 pushd cap-terraform/gke || exit
+git checkout -t origin/prabal-gke-clustername-tfvar
```
viccuad: We shouldn't have this.

Dockerfile
```diff
@@ -30,13 +30,10 @@ RUN curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.
...
   chmod +x aws-iam-authenticator && mv aws-iam-authenticator /usr/local/bin/
 
-RUN curl -o google-cloud-sdk.tar.gz https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-264.0.0-linux-x86_64.tar.gz && \
-  tar -xvf google-cloud-sdk.tar.gz && \
-  rm google-cloud-sdk.tar.gz && \
-  pushd google-cloud-sdk || exit && \
-  bash ./install.sh -q && \
-  popd || exit && \
-  echo "source /google-cloud-sdk/path.bash.inc" >> ~/.bashrc
+RUN curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz \
```
viccuad, line 33: :100:

backend/gke/deploy.sh
viccuad: Nevermind, I saw that the next commit removes it (it would be better to not have it in history for `git bisect`, etc)

Dockerfile
```diff
@@ -46,6 +43,11 @@ RUN curl -o install.py https://azurecliprod.blob.core.windows.net/install.py &&
   printf "\n\n\n\n" | python3 ./install.py && \
   rm ./install.py
 
+RUN helm_version=v3.1.1 \
```
viccuad, line 46: Nitpicking:
`helm` is not needed on the image neither for cluster creation nor for operation, the same way that `kubectl` or `cf-cli` aren't, since for now they get downloaded on `buildir/bin/` and pinned to the correct versions needed by the backend. So I would move this block under the `#extras` comment.

prabalsharma: I think I added this because of this 
```Installing CFO from: https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-3.3.0%2B0.gf32b521e.tgz
/tmp/build/b0d51b9f/catapult/include/func.sh: line 251: helm: command not found
make[1]: *** [Makefile:17: install] Error 127
make: *** [Makefile:182: scf] Error 2```
https://concourse.suse.dev/teams/main/pipelines/prabal-demo-release/jobs/deploy-diego-gke-sa/builds/7

viccuad: From the build, https://concourse.suse.dev/teams/main/pipelines/prabal-demo-release/jobs/deploy-diego-gke-sa/builds/7#L5e8e0bc6:13, which comes from https://github.com/SUSE/catapult/blob/540c4c8cfb4992c428bd64ab8c784f1b7a70ac26/modules/common/deps.sh#L23.

So maybe `DOWNLOAD_BINS=false` was incorrectly set in the pipeline, instead of using `DOWNLOAD_CATAPULT_DEPS=false`. 

Nevertheless, I think it's a good idea to add a helm version to the catapult image. But then, it's not a backend dep, therefore my nitpicking comment about the position of the paragraph.

PR #173, opened 2020-04-14
Use compute-optimized flavour instead of memory one for CaaSP

viccuad:
Use:
  c8.large: 	8 vcpus, 	8 GB ram,  160 GB disk
instead of:
  m1.xlarge: 	4 vcpus,	16 GB ram, 160 GB disk
  m1.xxlarge: 4 vcpus,	20 GB ram,	80 GB disk

### DIFF
--- a/backend/caasp4os/terraform-os/terraform.tfvars.skel
+++ b/backend/caasp4os/terraform-os/terraform.tfvars.skel
@@ -41,12 +41,12 @@ workers = 3
 # Size of the master nodes
 # EXAMPLE:
 # master_size = "m1.medium"
-master_size = "m1.xlarge"
+master_size = "c8.large"
 
 # Size of the worker nodes
 # EXAMPLE:
 # worker_size = "m1.medium"
-worker_size = "m1.xxlarge"
+worker_size = "c8.large"
 
 # Attach persistent volumes to workers
 workers_vol_enabled = 0


mudler:
There is also m8.xlarge on ECP just in case you need more ram (just checking, because I discovered it just few days ago :smiley:  )

HartS:
I think we do need more RAM after attempting to deploy 30-40 times with this flavour. I can't conclusively say that errors I'm seeing are tied to less RAM (8 GB in `c8.large` vs 16 GB in `m1.xlarge`), but I have seen kubecf deployments going into error state soon after being deployed, or failing to deploy. Additionally, I've recently been seeing a lot of CaaSP deployments which have failed with this:

```
Error: Error applying plan:

1 error(s) occurred:

* openstack_compute_instance_v2.master[2]: 1 error(s) occurred:

* openstack_compute_instance_v2.master.2: Error waiting for instance (dc9c925d-d4d5-4e0a-83b1-10fd0a6b5ab3) to become ready: unexpected state 'ERROR', wanted target 'ACTIVE'. last error: %!s(<nil>)
```

I'm not sure why VMs are going into an error state as there are no logs associated with the worker, and the only way to clean up after this happens is to go into the openstack console and manually delete the VM before running `make clean` (as this appears to be an issue with our version of the terraform provider)

viccuad:
This was superseded by https://github.com/SUSE/catapult/pull/218 , closing this. We can iterate more afterwards.

PR #178, opened 2020-04-16
Use kubecf's kube-ready-state-check.sh instead of scf's

viccuad:
And don't `|| true` anymore.

### DIFF
--- a/backend/check.sh
+++ b/backend/check.sh
@@ -3,13 +3,13 @@
 . include/common.sh
 . .envrc
 
-curl -s -Lo kube-ready-state-check.sh "$SCF_REPO"/raw/"$SCF_BRANCH"/bin/dev/kube-ready-state-check.sh
-chmod +x kube-ready-state-check.sh
-mv kube-ready-state-check.sh bin/
+curl -sSLo bin/kube-ready-state-check.sh \
+     https://raw.githubusercontent.com/cloudfoundry-incubator/kubecf/master/dev/kube/kube-ready-state-check.sh
+chmod +x bin/kube-ready-state-check.sh
 
-info "Testing imported k8s cluster"
+info "Testing imported k8s cluster for kubecf"
 
-kube-ready-state-check.sh kube || true
+kube-ready-state-check.sh kube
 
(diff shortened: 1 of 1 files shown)


backend/check.sh
```diff
@@ -3,13 +3,13 @@
 . include/common.sh
 . .envrc
 
-curl -s -Lo kube-ready-state-check.sh "$SCF_REPO"/raw/"$SCF_BRANCH"/bin/dev/kube-ready-state-check.sh
-chmod +x kube-ready-state-check.sh
-mv kube-ready-state-check.sh bin/
+curl -sSLo bin/kube-ready-state-check.sh \
+     https://raw.githubusercontent.com/cloudfoundry-incubator/kubecf/master/dev/kube/kube-ready-state-check.sh
```
viccuad, line 7: Hardcoding this instead of using SCF_REPO is painful.

But creating a new KUBECF_REPO is out of the scope for this PR, we really need to create `modules/kubecf/` at some point.

If we were to have KUBECF_REPO and SCF_REPO, we would need to have an `if kubecf then` at this point in time, when importing clusters, which also smells, as clusters shouldn't care about kubecf or scf.

We could run both kube-ready-state-check.sh, and only fail if both are failing (hence missing Tiller on kubecf will not be an issue).

mudler:
I guess it would be slightly better to run the checks after getting the chart, and actually move the check step as part of kubecf tear up.  In this way you can use the script included in the chart (if there is one) or the checkout.

Probably the backend needs a self-check just based on the kube state, not the requirements from the chart itself.

viccuad:
I opted for the non-hard link to get fixes from upstream for free. I figured either we do this the correct way, of having our own script, or just keep our benefits to the maximum by obtaining the check script on demand.

viccuad:
Closing this, it has conflicts after the separation into modules/kubecf and modules/scf.

By that separation, it was failing on env vars, commited fix https://github.com/SUSE/catapult/commit/3d08ca96c478050562cd39bff89f0468ea64d2aa.

We can revisit this with both the options provided in a different PR.

Issue #177, opened 2020-04-15
Ensure cluster setup and destroy can happen on different machines


prabalsharma:
~from https://github.com/SUSE/cap-terraform/tree/master/eks#instructions~

~Once you're done, destroy your infrastructure with terraform destroy. Note that sometimes, the internet gateway does not get detached and deleted even after 15 minutes and times out. If that happens, you'd have to manually delete the VPC and its dependent resources.~

~This is a blocker for catapult's use in CI.~

 ~eks clean script will either have to use aws cli to clean things or it can use https://github.com/colstrom/cloudstrap/blob/develop/bin/cloudstrap-teardown for easy teardown respecting dependencies.~

The original issue does not exist anymore so `terraform destroy` should be good enough for cleanup. However, there is the issue of the plan, apply and destroy potentially needing to work on different machines so there will need to be some work to make sure that can happen without problems as described in the comment below.

https://github.com/SUSE/catapult/issues/177#issuecomment-616230098

prabalsharma:
eksctl is a good way to delete stuff as AWS tests it to match their documentation. We won't have to maintain our own scripts to delete things

andreas-kupries:
@prabalsharma My questions so far below. I seem to have found answers for several as part of the research while writing them them up. Please still have a look and see if it needs any corrections. And if the questions ring bells about trouble spots to avoid, then knowing these would be nice too ;)

# Questions

With the cluster created by terraform, we will have some collection of
AWS resources.

  - How is that collection of resources mapped to something `eksctl` would understand ?

    I mean, if the cluster is created by `eksctl` it will

      1. know what resources are in the cluster, and will have
      2. stored relevant information in whtever config files it uses.

    With terraform, are your resources matching what `eksctl` does ?
  
    And how to translate/generate a suitable config `eksctl` can then
    use to tear that down ?

    __Answer__ I seem to have managed to answer my own question
    ... Clusters are referenced by name. So if I know the name of the
    cluster created by terraform I may have a chance of destroying
    it. At least what eksctl believes is part of it, vs what our
    terraform setup may create beyond that.

  - Currently looking at the "official" documentation at
    https://eksctl.io/ I am unable to find the cli reference, i.e. the
    full set of commands and their options.

    Ok. It seems you really have to download and install the cli
    locally so that you can use `eksctl help ...` for the commands and
    their options.

    Using that it seems a cluster is referenced by name.

  - Where in `catapult` is the EKS support ?

[...]

PR #182, opened 2020-04-17
Cleanup EKS cluster via eksctl

andreas-kupries:
References
- https://jira.suse.com/browse/CAP-1395
- https://github.com/SUSE/catapult/issues/177

Modifications
- Added `eksctl` to the set of installed dependencies/tools.
- Changed eks `clean.sh` to drop use of `terraform destroy` and use `eksctl delete cluster` instead.
  Cluster name is extracted from the `kubeconfig` file generated by terraform during setup.
  Brittle, relies on exact kubeconfig structure.
  Needs a `cap-terraform` PR adding this information as proper tf output.

:warning: :construction: To be tested :construction: :warning:

### DIFF
--- a/modules/common/deps.sh
+++ b/modules/common/deps.sh
@@ -23,7 +23,7 @@ fi
 if [[ "$DOWNLOAD_BINS" == "false" ]]; then
     ok "Skipping downloading bins, using host binaries"
 else
-    info "Downloading specific helm, kubectl, cf versions…"
+    info "Downloading specific helm, kubectl, cf, eksctl versions…"
     if [ ! -e "bin/helm" ]; then
         curl -sSL https://get.helm.sh/helm-${HELM_VERSION}-${HELM_OS_TYPE}.tar.gz | tar zxf -
         mv $HELM_OS_TYPE/helm bin/
@@ -62,7 +62,6 @@ else
         eksctl version
     fi
 fi
-
 
 if [[ "$DOWNLOAD_CATAPULT_DEPS" == "false" ]]; then
     ok "Skipping downloading catapult dependencies, using host binaries"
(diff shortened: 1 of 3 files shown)


modules/common/deps.sh
```diff
@@ -51,8 +51,17 @@ else
         info "CF cli version:"
         cf version
     fi
-fi
 
+    if [ ! -e "bin/eksctl" ]; then
```
prabalsharma, line 55: do not forget Dockerfile

andreas-kupries: There is no dockerfile in the eksctl tarball.
... :bulb: You are talking about a Dockerfile in catapult ?!
... I see, top dir `Dockerfile`.
Will do.

andreas-kupries: Done with commit c03441fd40b05f9dce399c1e71f9bfed43b080cb

modules/common/deps.sh
```diff
@@ -51,8 +51,17 @@ else
...
+	curl --silent --location \
```
viccuad, line 56: This should go into https://github.com/SUSE/catapult/blob/master/backend/eks/deps.sh, that way it doesn't get downloaded for all backends, and the download is skipped with `DOWNLOAD_CATAPULT_DEPS=false` as in CI and catapult unit/integration tests.

Also, would you like to add a couple of lines in `catapult/tests/{unit,integration}-tests.sh` for `eksctl` :D? One can look at the existing ones grepping for `DOWNLOAD_CATAPULT_DEPS` there.

andreas-kupries: I will try to remember for future works on new tools needed by catapult.

andreas-kupries:
Closing this PR as irrelevant. The original issue motivating looking into use of `eksctl` (terraform destroy not fully cleaning up) seems to be gone, and the attempted mix&match is not working out. IOW this change here is a dead end we have to abandon.

Some more notes in #177 ticket.

PR #184, opened 2020-04-20
Delete kubecf chart bits when make scf-clean

viccuad:


mudler:
Given https://github.com/SUSE/catapult/blob/f7b04f437ae3e0bc0203049a0c2b94875b17aab3/include/common.sh#L37 , shouldn't we remove only if file exists? (or everything will just fail?)

viccuad:
With --force on `rm`, it ignores nonexistent files, so it doesn't fail.

viccuad:
Added more files to be deleted. Ideally we should put everything from the chart inside a specific folder, or redownload/reuse the tar to list the files to be deleted.

mudler:
> With --force on `rm`, it ignores nonexistent files, so it doesn't fail.

missed that, sorry :relaxed:

viccuad:
Merging, thanks!

Issue #185, opened 2020-04-20
Catapult needs to set DNS via cloud DNS service for hosted kubernetes providers


HartS:
When LoadBalanced type services are used on hosted kubernetes environments (confirmed with AKS, and expected with GKE and EKS as well), the kubecf, stratos, and metrics deployments need to confirm that public DNS records have been set appropriately before considering a deployment successful.

Previously with cf-ci we scripted a wait 

and used `az network dns` [to wait for loadbalancer services to become ready and then set dns records set records](https://github.com/SUSE/cf-ci/blob/develop/qa-pipelines/tasks/cf-deploy.sh#L57-L58) (regardless of the actual kube provider)

However, we would probably benefit from using something like [external-dns](https://github.com/kubernetes-sigs/external-dns) which would unify the steps for the kubernetes providers we support/test while using their associated dns services

viccuad:
Closing this, since #192 achieved it.

PR #192, opened 2020-04-27
Add SCF_SERVICES, use external-dns for DNS on GKE

viccuad:
This PR:

* Adds option `SCF_SERVICES={lb,ingress}` (default `lb`). With that:
  - Constructs scf-config-values.yaml correctly depending on SCF_SERVICES
  - Annotates kube services as needed so the `external-dns` helm release takes care of creating DNS entries on the current public cloud provider
  - Deletes annotations on `make clean` or `make scf-clean` correctly
* Consumes new changes on cap-terraform/gke to be able to test the new `SCF_SERVICES` option. Adds a `GKE_DNSDOMAIN` option for that effect.


viccuad:
Tested deploying a kind cluster with lb (the default deployment), works:
```
SCF_OPERATOR=true ENABLE_EIRINI=false SCF_SERVICES=lb make k8s scf scf-login sample
```

Tested deploying a GKE cluster with LB and automatic DNS entries, works:
```
BACKEND=gke CLUSTER_NAME=gke-lb \
  GKE_CRED_JSON=<>.json \
  SCF_OPERATOR=true ENABLE_EIRINI=false \
  SCF_SERVICES=lb \
  make k8s scf scf-login sample
```

Testing deploying a GKE with Ingress and automatic DNS entries:
```
BACKEND=gke CLUSTER_NAME=gke-ingress \
  GKE_CRED_JSON=<>.json \
  SCF_OPERATOR=true ENABLE_EIRINI=false \
  SCF_SERVICES=ingress \
  make k8s scf scf-login sample
```

TODO:

* unit tests fail because calls to `make clean` or `make scf-clean` (both call `modules/scf/clean.sh` now because of the annotations) expect a kubeconfig.
* concerns about time needed for external-dns to delete the DNS entries (see external_dns_* functions): make sure deleting services delete dns entries
* Use a parametrized DNS zone that allows for *clustername*.ci.foo

Makefile
```diff
@@ -22,6 +22,7 @@ buildir:
 # General targets (Public)
 .PHONY: clean
 clean: ##@STATES Delete cluster of type $BACKEND and location build$CLUSTER_NAME
+clean: scf-clean # first scf-clean because of external-dns annotations
```
andreas-kupries, line 25: Just a side note: This is also needed (at least for AWS EKS) to ensure that the kubecf PVC are deleted when tearing things down. Just doing `clean` leaves them active if unused, and they keep accumulating costs. So +1 on this piece of the whole.

andreas-kupries:
I do not understand the notation `GKE_CRED_JSON=<>.json` in the example code. what does `<>` stand for ?

andreas-kupries:
Trialing with minikube: ```BACKEND=minikube CLUSTER_NAME=_dns_trial SCF_OPERATOR=true SCF_SERVICES=lb make k8s scf scf-login sample```

ended in ```   2020/04/28 00:04:47 Building droplet release
   2020/04/28 00:04:47 Creating app artifact

Waiting for app to start...
Start app timeout

TIP: Application must be listening on the right port. Instead of hard coding the port, use the $PORT environment variable.

Use 'cf logs cf-nodejs --recent' for more information
FAILED
```
Hm. Looking at services ```kubectl get svc --namespace scf
NAME                    TYPE           CLUSTER-IP       EXTERNAL-IP                   PORT(S)                                                                                                                                                        AGE
bits                    LoadBalancer   10.97.167.169    <pending>                     6666:32467/TCP                                                                                                                                                 65m
eirini-registry         NodePort       10.101.226.229   <none>                        6666:31666/TCP                                                                                                                                                 65m
router-public           LoadBalancer   10.97.203.161    192.168.39.99,192.168.39.99   80:32635/TCP,443:32179/TCP                                                                                                                                     65m
ssh-proxy-public        LoadBalancer   10.99.58.34      192.168.39.99,192.168.39.99   2222:32059/TCP                                                                                                                                                 65m
tcp-router-public       LoadBalancer   10.110.195.164   192.168.39.99,192.168.39.99   80:30649/TCP,20000:30456/TCP,20001:30737/TCP,20002:31505/TCP,20003:30700/TCP,20004:31102/TCP,20005:31656/TCP,20006:32038/TCP,20007:31070/TCP,20008:31302/TCP   65m
```

Note : First attempt using diego instead of eirini was aborted manually after it became clear that diego was not coming up (CLBO it was not recovering from).

No annotations ...

andreas-kupries:
Right. Being minikube, and the code for annotations is just in GKE. So not a surprise, and not working might also be outside of this ticket.

HartS:
> A +1 on the make clean/scf-clean linkage.
> A nit on the double negation thing.
> 
> You said in the 🚀 that `It provides the plumbing for consuming terraform files for EKS, AKS `.
> That part I simply do not see in the PR. I suspect my inexperience with catapult.
> WIBNI if you could point me directly to the parts which facilitate that. Here or via code comments.
> 
> See also my PR #193 which feels to me to be more what is needed in terms of code (and is quite possibly wrong, definitely limited to EKS).
> 
> Also, title says `public clouds` (i.e. general, implies all), code deals only with GKE, nothing seen about AKS, nor EKS. Planned to be in a separate PR ? Or should it be added here before merging ?

@andreas-kupries I *think* (hope) external-dns works correctly for the cloud provider it's installed on automatically. Each of these platforms would certainly be good to test this work on though.

If this indeed works on AKS, EKS, GKE, it closes https://github.com/SUSE/catapult/issues/185

Makefile
viccuad: True, reworded comment :)

include/func.sh
```diff
@@ -289,3 +289,39 @@ function wait_for {
...
+    kubectl annotate svc uaa-uaa-public \
+            -n "$ns" \
+            "external-dns.alpha.kubernetes.io/hostname=uaa.${domain}, *.uaa.${domain}"
+}
+
+
+external_dns_annotate_scf() {
+    local ns=$1;shift
+    local domain=$1;shift
+    if [ "${SCF_OPERATOR}" != "true" ]; then
```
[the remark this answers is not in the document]
viccuad: Thanks! undone double negation

viccuad:
Thanks for the review!

> You said in the rocket that It provides the plumbing for consuming terraform files for EKS, AKS .
That part I simply do not see in the PR. I suspect my inexperience with catapult.
WIBNI if you could point me directly to the parts which facilitate that. Here or via code comments.
> See also my PR #193 which feels to me to be more what is needed

You are right, I thought that I had time to include everything here when it was marked as WIP. Reworded title more accurately.

> I do not understand the notation GKE_CRED_JSON=<>.json in the example code

It's just the path of your creds json, I hastily edited it when commenting. 

For minikube, it should be possible to leave the `$DOMAIN` as `external-ip-of-1-node.omg.howdoi.website`, and it should work. I haven't tried minikube with kubecf though, it may be that it's not up to date. It would be good to take ideas from bazel's kubecf there.

I'm using kind for testing non-public clouds.

viccuad:
I consider this merge-able, and we can figure out the ingress side later. It may be done in the scf state side, like https://github.com/suse/catapult/commit/aa3855b, or better, yet, leave it as it is, and install ingress only once on the backend side.

backend/gke/deploy.sh
```diff
@@ -31,7 +31,10 @@ node_count     = "$GKE_NODE_COUNT"
 vm_type        = "UBUNTU"
 gke_sa_key     = "$GKE_CRED_JSON"
 gcp_dns_sa_key = "$GKE_CRED_JSON"
-cluster_labels = {key = "$GKE_CLUSTER_NAME"}
+cluster_labels = {
+    catapult-clustername = "$GKE_CLUSTER_NAME",
```
prabalsharma, line 35: whats going on here? 
https://github.com/SUSE/cap-terraform/blob/master/gke/terraform.tfvars.template#L8 : states just key = value. 
may be a comment or something

backend/gke/deploy.sh
```diff
@@ -56,12 +59,11 @@ info "Configuring deployed GKE cluster…"
 ROOTFS=overlay-xfs
 # take first worker node as public ip:
 PUBLIC_IP="$(kubectl get nodes -o json | jq -r '.items[].status.addresses[] | select(.type == "InternalIP").address' | head -n 1)"
-DOMAIN="$PUBLIC_IP.$MAGICDNS"
 if ! kubectl get configmap -n kube-system 2>/dev/null | grep -qi cap-values; then
     kubectl create configmap -n kube-system cap-values \
             --from-literal=garden-rootfs-driver="${ROOTFS}" \
             --from-literal=public-ip="${PUBLIC_IP}" \
```
prabalsharma, line 65: what is the use of public IP? when we will be using lb svc data to create dns

backend/gke/deploy.sh
satadruroy: yeah, the template file just has placeholders. The proposed labels here look fine to me.

andreas-kupries: `key = value` is the general syntax. Importantly, `key` can be any string.
In the old code the key was named `key`, which did not say much.
Now the key is called the more descriptive `catapult-clustername`.
This goes into the cluster label section of the AWS console.
IIRC this also allows search by key and/value in the console.
Makes finding the clusters and cleaning up after trouble easier.

backend/gke/deploy.sh
andreas-kupries: I do not believe keeping it for now will hurt. When we know for sure to not need it in this setup we can remove it.

modules/scf/gen_config.sh
```diff
@@ -18,6 +18,8 @@ if [ "$SCF_OPERATOR" != true ]; then
     fi
 fi
 
+kubectl patch -n kube-system configmap cap-values -p $'data:\n services: "'$SCF_SERVICES'"'
+services="$SCF_SERVICES"
 domain=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["domain"]')
 public_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["public-ip"]')
```
prabalsharma, line 24: again, Public ip should not be required for public clouds, AFAIK

modules/scf/defaults.sh
```diff
@@ -13,6 +13,7 @@ OPERATOR_CHART_URL="${OPERATOR_CHART_URL:-latest}"
 
 # scf-gen-config relevant:
 
+SCF_SERVICES="${SCF_SERVICES:-lb}" # lb, ingress
```
prabalsharma, line 16: I was under impression that lb is used even for ingress. You can do lb without ingress but not ingress without lb. I might be wrong.

backend/gke/deploy.sh
prabalsharma: No, that is this work. To remove public ip and domain combo with dns entries using lb service.

backend/gke/deploy.sh
prabalsharma: ok. But when CI will use this; `owner = "$(whoami)".` it will provide a value `root`. Which won't make much sense. 
It was better to keep it `{key = "$GKE_CLUSTER_NAME"}`, as that would allow local deployment and CI to use relevant label.
For local deployment if someone don't want to provide a relevant label, then it anyways defaults to` whoami` : `GKE_CLUSTER_NAME="${GKE_CLUSTER_NAME:-$(whoami)-cap}"`

modules/scf/gen_config.sh
```diff
@@ -56,12 +58,27 @@ fi
 
 if [ "${SCF_OPERATOR}" == "true" ]; then
 
+
+if [ "$services" == ingress ]; then
+INGRESS_BLOCK="ingress:
```
prabalsharma, line 63: https://github.com/SUSE/catapult/pull/192#issuecomment-620717537, if ingress is not ready yet. Then we should exit here.

include/func.sh
```diff
@@ -289,3 +289,41 @@ function wait_for {
     timeout=300
     n=0; until ((n >= timeout)); do eval "$1" && break; n=$((n + 1)); sleep 1; done; ((n < timeout))
 }
+
+external_dns_annotate_uaa() {
+    local ns=$1;shift
+    local domain=$1;shift
+    kubectl annotate svc uaa-uaa-public \
+            -n "$ns" \
+            "external-dns.alpha.kubernetes.io/hostname=uaa.${domain}, *.uaa.${domain}"
```
prabalsharma, line 298: how will it handle parallel deployments? won't we need to use a sud domain? ` uaa.${cluster name}.${domain}`

andreas-kupries: AFAICT the domain we see here comes from the `GKE_DNSDOMAIN`. While that has a default it can be overridden by the user. So I guess the pipeline using catapult could do that, inserting configuration data (pipeline name, diego/eirini, sa/ha), and time-dependent data. The latter would be what would allow parallel deployments.

backend/gke/deploy.sh
viccuad: I took the `owner` label idea from Satadru's clusters on GKE, there were more visible than the cluster names. Also, Since the cluster name can be changed, I find value in having a label that is just owner and cannot be changed. Simplifies searching for your clusters.

include/func.sh
viccuad: That's the idea, for the user/ci to pass a domain that can be used. I tested deploying clusters as `*.gke-lb.ci.kubecf.charmedquarks.me` and `gke-ingress.ci.kubecf.charmedquarks.me`, but with that specific domain did not work.
It should be possible to create a domain that works, and if not, we may need to just have a list of empty domains for grabs.

backend/gke/deploy.sh
viccuad: I think there's value on having the ip's so they get added by default on the NodePorts. it doesn't hurt, and we need a public ip anyways for kind & minikube, until we migrate them to something else.

modules/scf/gen_config.sh
viccuad: (see other comment)

modules/scf/defaults.sh
viccuad: AFAIU Ingress depends on the k8s provider, so it can be that you get an ingress termination without a LB.

:thinking:  maybe `SCF_SERVICES` is a misnomer, but I can't find anything better. It's an option to specify how the services communicate externally. There's 3 possibilities: nodeport (ip, as we did, but not useful for kubecf), lb (k8s provider LBs, what kubecf defaults to), ingress (k8s provider dependent, on GKE it actually spawns a GKE LB).

modules/scf/gen_config.sh
viccuad: Thing is, ingress configuration seems to be ready, I have compared it with tf3-c4.cap.suse.de ingress setup.

I can't get it to work on GKE. I get:
`Request error: Get https://api.ci.kubecf.charmedquarks.me: dial tcp 34.<edited>:443: connect: connection refused`

I get correct ingresses:
```
vic@viccuad ~/suse/catapult-aks/buildgke-lb$ kubectl describe ingress susecf-scf -n scf
Name:             susecf-scf
Namespace:        scf
Address:          34.<edited>
Default backend:  default-http-backend:80 (<none>)
TLS:
  susecf-scf-ingress-tls terminates *.ci.kubecf.charmedquarks.me,ci.kubecf.charmedquarks.me
Rules:
  Host                          Path  Backends
  ----                          ----  --------
  *.ci.kubecf.charmedquarks.me
                                /   router:443 (10.<edited>:443)
  ci.kubecf.charmedquarks.me
                                /   router:443 (10.<edited>:443)
Annotations:
  nginx.ingress.kubernetes.io/secure-backends:   true
  nginx.ingress.kubernetes.io/ssl-redirect:      false
  nginx.org/websocket-services:                  router
  kubernetes.io/ingress.class:                   nginx
  nginx.ingress.kubernetes.io/backend-protocol:  HTTPS
  nginx.ingress.kubernetes.io/proxy-body-size:   8m
Events:                                          <none>
```
And the router service seems to be ok:
```
NAME                    TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                                                                            AGE                                            
router                  ClusterIP   None            <none>        80/TCP,443/TCP                                                                                     20m                                            
router-0                ClusterIP   10.<edited>   <none>        80/TCP,443/TCP                                                                                     20m                                                
```

I'm resolving the conversation, I believe the ingress code is ok.

prabalsharma:
@viccuad I still think that the dns entires should use subdomain of GKE_DNSDOMAIN. I will implement that in my work. Merging this for now.

viccuad:
@prabalsharma Yep, I agree. maybe it would have been better to leave the GKE_DNSDOMAIN undefined by default, to force users to provide one.

PR #195, opened 2020-05-04
Consume cap-ci branch from cap-terraform

viccuad:
`cap-ci` branch contains the following fixes:
- `controller.publishService.enabled` on deployment of external-dns for GKE, which fixes creating DNS entries when using ingress
- set a unique `txtOwnerId` in the TXT record for the external-dns chart, which solves DNS entries getting overwritten by different clusters.

PR #196, opened 2020-05-04
Add aks skeleton

viccuad:
Add `aks/deploy.sh`, and `aks/clean.sh` with tentative configuration for terraform vars.


viccuad:
As it is it may be incomplete because of the DNS principal setup. Nevertheless, I would like to get this merged, to not start diverging from https://github.com/SUSE/catapult/pull/195.

PR #200, opened 2020-05-05
tf s3 backend for gke

prabalsharma:
- tf s3 backend for GKE
- tf state file (if provided) for tf destroy
- subdomain for gke dns


backend/gke/clean.sh
```diff
@@ -11,6 +12,16 @@ set -exuo pipefail
 if [ -d "$BUILD_DIR" ]; then
     . .envrc
 
+    if [ -f "$TFSTATE" ]; then
+        mkdir -p cap-terraform/gke
+        (cd cap-terraform/gke || exit
+          # ATTENTION: The next command overwrites existing files without
```
bikramnehra, line 18: can you consolidate the comment in single line?

backend/gke/deploy.sh
```diff
@@ -39,16 +39,33 @@ cluster_name   = "$GKE_CLUSTER_NAME"
...
 
 info "Deploying GKE cluster with terraform…"
 
 terraform init
 
-terraform plan -out="$(pwd)"/my-plan
+terraform plan -out=my-plan
+
+if [ -n "${TF_KEY}" ] ; then
+    zip -r9  "${BUILD_DIR}/tf-setup.zip" .
```
bikramnehra, line 64: I am not sure exactly what's being done here.

prabalsharma: zips the file for concourse pool. Same as in `master` for EKS: https://github.com/SUSE/catapult/blob/master/backend/eks/deploy.sh#L50-L51

prabalsharma:
> Overall I don't like how we are rendering TF templates but I guess its a design issue and shouldn't be solved in this PR. Maybe we can use `gomplate` or something like that.

Yes, the design is already approved and  in `master `  https://github.com/SUSE/catapult/blob/master/backend/eks/deploy.sh#L35-L41.

PR #209, opened 2020-05-13
Fix security group handling for caasp4os backend

dirkmueller:
There were several bugs in this code:
* it was using a mix of nova compute and neutron networking security
  groups, that is not very stable and can cause races. use networking
  rules consistently.
* it was using 'name' for assigning security groups, which fails
  if there are duplicates. better use 'id's as those are ids
* simplify rules a bit
* fix missing cap-rules assignment for accessing cf endpoint.


viccuad:
Many thanks!
Tested by successfully creating a cluster, kubecf deployment ongoing.
Merging already.

PR #211, opened 2020-05-15
Add kubecf module, separate kubecf and scf code, provide compat

viccuad:
This PR separates and simplifies kubecf and scf code without breaking the UI.

It:
- Adds a new `module/kubecf/*`, and its targets `make kubecf*`.
- Provides compatibility for users & ci by pointing `SCF_OPERATOR make scf*` to `make kubecf`.
- Maintains the rest of `SCF_*` variables for both modules, to not break compatibility.
- Refactors kubecf code into `modules/kubecf/*`, and scf code into `modules/scf/*`.
- Removes the need of SCF_OPERATOR on both modules.
- Adds sane defaults for both kubecf and scf modules.

It maybe it's useful to review this PR by commits instead of files, given the copied files, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -208,6 +208,19 @@ kubecf-build-stemcell: ##@kubecf Build stemcell for KubeCF
 	$(MAKE) -C modules/kubecf stemcell_build
 
 # scf-only targets:
+ifeq "$(SCF_OPERATOR)" "true"
+scf-build: kubecf-build
+scf-clean: kubecf-clean
+scf: kubecf
+scf-chart: kubecf-chart
+scf-gen-config: kubecf-gen-config
+scf-install: kubecf-install
+scf-upgrade: kubecf-upgrade
+scf-login: kubecf-login
+scf-minibroker: kubecf-minibroker
+scf-purge: kubecf-purge
+scf-build-stemcell: kubecf-build-stemcell
+else
 .PHONY: scf-build
 scf-build: ##@scf Build chart from source and install CF
 	$(MAKE) -C modules/scf build-scf-from-source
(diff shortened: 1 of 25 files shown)


viccuad:
I suppose we should inform people to target their fixes in `modules/kubecf` and not `modules/scf` (as there's no kubecf code anymore in the latter!) at the same time we merge this.

Makefile
```diff
@@ -160,14 +160,67 @@ module-experimental-eirinifs:
...
+.PHONY: kubecf-purge
+kubecf-purge: ##@kubecf Purge all apps, buildpacks and services from KubeCF
+	$(MAKE) -C modules/kubecf purge
+
+.PHONY: kubecf-build-stemcell
+kubecf-build-stemcell: ##@kubecf Build stemcell for KubeCF
+	$(MAKE) -C modules/kubecf stemcell_build
+
+# scf-only targets:
+ifeq "$(SCF_OPERATOR)" "true"
```
mudler, line 211: Maybe enhance the comment a bit, so when we will look at this after few weeks, we will remember why we did this :slightly_smiling_face:

mudler:
I've created the [v1.1.0](https://github.com/SUSE/catapult/releases/tag/v1.1.0) release, so in case we can tag those in the pipelines - just in case

Makefile
viccuad: True! many thanks, fixed.

viccuad:
Tested successfully by doing `make all` (which now defaults to kubecf), `make kubecf`, `make scf`, `SCF_OPERATOR=true make scf` (which deploys kubecf for compat reasons, and it's what it is run on the CI right now).

`make scf-clean` seems broken, but it's the same without this PR, so it's not a regression.

PR #212, opened 2020-05-18
Reduce docker image size

viccuad:
Reduced from 3.12GB to 2.88GB.
Achieved by:
- Reducing the default components for gcloud
- Removing unneeded locales

Successfully tested by spawning the image, and trying to login and list account settings with `gcloud` and `az`.


viccuad:
Note that merging this may fix the Concourse pod workers problem with volumes, which may give red herrings to people trying to fix that.

PR #216, opened 2020-05-19
Add corenumber option for CATS_NODES

viccuad:
if CATS_NODES set to "corenumber", spawn as many ginkgo nodes as cores the
k8s workers have.

### DIFF
--- a/modules/kubecf/gen_config.sh
+++ b/modules/kubecf/gen_config.sh
@@ -32,10 +32,10 @@ else
 INGRESS_BLOCK=''
 fi
 
-last_worker=$(kubectl get nodes | tail -1 | cut -f 1 -d ' ')
-num_cores=$(kubectl describe node "$last_worker" | \
-          grep Allocatable -A 2 | grep cpu | tr -s ' ' | cut -f 3 -d ' ')
 if [ "$CATS_NODES" == corenumber ]; then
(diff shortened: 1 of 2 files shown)


modules/kubecf/gen_config.sh
```diff
@@ -32,6 +32,13 @@ else
 INGRESS_BLOCK=''
 fi
 
+last_worker=$(kubectl get nodes | tail -1 | cut -f 1 -d ' ')
+num_cores=$(kubectl describe node "$last_worker" | \
+          grep Allocatable -A 2 | grep cpu | tr -s ' ' | cut -f 3 -d ' ')
+if [ "$CATS_NODES" == corenumber ]; then
```
jimmykarily, line 38: [MINOR] Maybe move the 2 lines above inside the `if`? This will save some time from describing and getting the nodes when it's not needed.

viccuad: Doh, true! done.

viccuad:
> LGTM. We might want to propagate this over to cap-ci pipelines as well.

I'm conflicted now to set `corenumber` and discovery as default. KVM based CaaSP machines for example have 15 vcpus per worker node, but those threads are shared between other VM nodes.

We may need to default to `1`, and use `corenumber` as a per-backend setting on the pipelines.

viccuad:
Closing because we are never sure of the number of cores we need, and it needs manual interaction.

PR #217, opened 2020-05-19
Make aks deployment work

greygoo:


backend/aks/clean.sh
```diff
@@ -9,6 +9,9 @@ if [ -d "$BUILD_DIR" ]; then
     . .envrc
 
     if [ -d "cap-terraform/aks" ]; then
+
+	export KUBECONFIG="$(pwd)"/cap-terraform/aks/aksk8scfg
```
viccuad, line 13: Seems that after last lint fix, shellcheck SC2155 appears. This ought to fix it:

```suggestion
	KUBECONFIG="$(pwd)"/cap-terraform/aks/aksk8scfg
	export KUBECONFIG
```

greygoo: Thanks, applied the change.

viccuad:
merging! :D

PR #218, opened 2020-05-20
Rightsize the flavors a bit.

dirkmueller:
The master nodes I think don't need to be icnreased. and
for the worker nodes m1.xlarge provides more disk and vcpus actually.


viccuad:
Thanks, merging. For the masters, it's totally fair. And I forgot about the resizing of m1.xlarge; so that fits the workers better nowadays.

PR #224, opened 2020-06-02
Let the user explicitly specify the DOMAIN.

jimmykarily:
Can be used when importing an existing cluster from public cloud for
which proper DNS is setup.


jimmykarily:
I don't know if the same override mechanism should be implemented in other places too.

PR #225, opened 2020-06-02
Adding changes for AKS deploy and clean.

bikramnehra:
This PR adds changes related to AKS deploy and clean step.


backend/aks/deploy.sh
```diff
@@ -77,7 +78,7 @@ kubectl get svc
 ROOTFS=overlay-xfs
 # take first worker node as public ip:
 wait_for 'PUBLIC_IP="$(kubectl get services nginx-ingress-nginx-ingress-controller -o json | jq -r '.status[].ingress[].ip' 2>/dev/null)"'
-DOMAIN="$CLUSTER_NAME.$MAGICDNS"
+DOMAIN="$AZURE_CLUSTER_NAME.$MAGICDNS"
```
prabalsharma, line 81: looks like AKS work still is not using LB. This won't work.

satadruroy: > looks like AKS work still is not using LB. This won't work.

You are correct that this seems to make an assumption that the subsequent KubeCF deployment happens with an Ingress Controller but if KubeCF is deployed with plain LB services and no Ingress Controller then this external-ip will have no effect.

However, things should still work as the external ip set in the `cap-values` configmap will have no effect when the `external-dns` annotations are applied because `external-dns` does not rely on any of our external ips - it just knows the service names it has to create DNS entries for and it looks up those external IPs corresponding to those services to do the mapping.

bikramnehra: for comparison, we are simply using node's `InternalIP` here for `GKE` deployment:
https://github.com/SUSE/catapult/blob/master/backend/gke/deploy.sh#L89

prabalsharma:
LGTM

backend/aks/deploy.sh
prabalsharma: we are not using `InternalIP`. It was just not removed. 
 Proper dns implementation for AKS can be handled in a separate PR.

PR #227, opened 2020-06-04
Adding required env vars for Azure SP.

bikramnehra:


backend/aks/defaults.sh
```diff
@@ -3,7 +3,7 @@
 # AKS options
 ##################
 
-KUBECTL_VERSION="${KUBECTL_VERSION:-1.18.2}"
+KUBECTL_VERSION="${KUBECTL_VERSION:-v1.18.2}"
```
viccuad, line 6: :100:

backend/aks/deploy.sh
```diff
@@ -14,6 +14,8 @@ if ! az account show; then
        --username "${AZURE_APP_ID}" \
        --password "${AZURE_PASSWORD}" \
        --tenant "${AZURE_TENANT_ID}"
+
+    az account set --subscription="${AZURE_SUBSCRIPTION_ID}"
```
viccuad, line 18: Nitpicking: I would then change the print of `info "Missing azure credentials, running az login…"` to `login into the account` or so.

backend/aks/deploy.sh
```diff
@@ -33,6 +35,9 @@ helm_init_client
 
 # ssh_public_key needs to be a file. Build it regardless of {ssh,gpg}-agent, or
 # forwarding of agents:
+eval "$(ssh-agent)"
```
viccuad, line 38: I'm not sure of this.
The idea was to consume whatever key was in the keychain. That way, your personal clusters are always accessible with your key, and the CI ones with a (yet to be created) shared key.

I would prefer to fail with a message on `ssh-add -L`, if there are no keys on the keychain, instead of always creating one.

Also, note the comment just before this line: it used to take whatever ssh key was in the keychain, regardless if the keychain is managed by ssh-agent or gpg-agent. Now, it will always use ssh-agent.

viccuad:
This may need an update of the wiki https://github.com/SUSE/catapult/wiki/AKS

backend/aks/deploy.sh
bikramnehra: done.

backend/aks/deploy.sh
bikramnehra: created a new secret `ssh-key-ci` on concourse.
corresponding code in `cap-ci`: https://github.com/SUSE/cap/pull/28/files#diff-ddacb7f6f728e8d3719cd265216c9a1bR18

PR #229, opened 2020-06-05
Fix kubecf minibroker

viccuad:
Update `modules/kubecf/minibroker.sh` for helm3 and add missing suse helm repo.
Delete minibroker parts if present, in `modules/kubecf/clean.sh`


viccuad:
merging, doesn't touch anything out of it, and right now it's broken.

PR #230, opened 2020-06-05
Call {stratos,metrics}-clean on kubecf-clean

viccuad:
If kubecf is gone, any stratos install would be broken too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,8 @@ kubecf-build: ##@kubecf Build chart from source and install KubeCF
 	$(MAKE) -C modules/kubecf install
 
 .PHONY: kubecf-clean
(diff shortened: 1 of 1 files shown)


viccuad:
What do you folks think?
I think we need something similar, question is, if adding it to `make kubecf-clean` (such as here) or add a new `make kubecf-clobber` or so, with the same functionality (remove kubecf, stratos, metrics, etc).

bikramnehra:
> What do you folks think?
> I think we need something similar, question is, if adding it to `make kubecf-clean` (such as here) or add a new `make kubecf-clobber` or so, with the same functionality (remove kubecf, stratos, metrics, etc).

I am fine either way. But it would be great to have less things to take care of.

PR #232, opened 2020-06-12
Cleanup testing pods before starting a new run

jimmykarily:
quarks job deletes the generated job but doesn't delete the pod unless
the job has the lable "delete: pod".

This could be achieved by this:
https://github.com/cloudfoundry-incubator/kubecf/pull/1015

but that could delete the pod too fast before we get to write the logs
in a file. Also, keeping the pod around can be useful for debugging.


viccuad:
I've confirmed that helm upgrade doesn't restart the test pods and the finished job (I remember it doing it in the past, so, good).

PR #234, opened 2020-06-16
Don't hardcode EKS instance_type, use terraform's

viccuad:
Needs https://github.com/SUSE/cap-terraform/pull/74


viccuad:
https://github.com/SUSE/cap-terraform/pull/74#event-3449471772 is merged, merging this then

PR #239, opened 2020-06-23
[caasp4os] Enable security group management in CPI config

stefannica:
When the OpenStack cloud provider is enabled, all cloud load
balancer instances created to reflect LoadBalancer type services
are automatically associated with the default security group in
the OpenStack project specified in the cloud provider configuration.
There is a cloud provider setting, manage-security-groups [1],
that can be enabled to allow the kubernetes OpenStack cloud provider
to manage load balancer security group rules automatically to match
the service ports.

The node-security-group setting has been deprecated [2] and is not
required anymore.

[1] kubernetes/kubernetes#31921
[2] kubernetes/kubernetes#58145

Signed-off-by: Stefan Nica <snica@suse.com>

### DIFF
--- a/backend/caasp4os/terraform-os/generate-cpi-conf.sh
+++ b/backend/caasp4os/terraform-os/generate-cpi-conf.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#shellcheck disable=SC2145,SC2016
+log()   { (>&1 echo -e "$@") ; }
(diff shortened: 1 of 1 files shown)


viccuad:
Disregard my comment, tested by Stefan, it works. Merging

stefannica:
> Is this shellscript actually going to get executed in catapult?

This script is invoked by a terraform null resource to generate the cloud provider configuration.

PR #240, opened 2020-06-25
fix: kubecf: quarks-operator 5.x compatibility

mook-as:
Make the cf-operator install command support both 4.x and 5.x (by supplying the required namespace both ways).

This was used to push operator 5.0.0 to GKE; seems to work.

### DIFF
--- a/include/func.sh
+++ b/include/func.sh
@@ -284,6 +284,12 @@ function helm_ls {
     fi
 }
 
+# Given a chart path in $1, find its app version
+function helm_chart_app_version {
+    # helm 3 aliases "inspect" to "show", so the syntax is the same for 2 & 3.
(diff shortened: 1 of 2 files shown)


mook-as:
Amended to detect based on the operator version. `helm_ls` wasn't appropriate in this case (it looks at already-deployed charts), so I added a function to print out the app version of an uninstalled chart instead.

viccuad:
more verbose that I expected, to be honest, but at least it lends better to refactoring.

It feels that new catapult approaches should have a deployer per helm chart (one for cf-operator, another for kubecf, another for stratos..). And those deployers should minimally patch a default values.yaml from the chart. We keep growing weird one-offs..

PR #242, opened 2020-06-26
Fix gke/kubeconfig.sh, we need to source scripts before anything

viccuad:
If not:
- call to `err` in line 10 will fail
- we will not be inside of buildfolder, and we will lack the binary for yq

### DIFF
--- a/backend/gke/kubeconfig.sh
+++ b/backend/gke/kubeconfig.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+. ./defaults.sh
+. ../../include/common.sh
+. .envrc
+
 # KUBECFG can be a kubeconfig file or a cluster reference file for local deployments
 # KUBECFG has to be a kubeclusterreference for CI usage, check gke/deploy.sh for format
 if [ ! -f "$KUBECFG" ]; then
@@ -20,10 +24,6 @@ else
     echo "Please check your KUBECFG"
     exit 1
 fi
-
-. ./defaults.sh
-. ../../include/common.sh
-. .envrc
 
 # check gcloud credentials:
 info "Using creds from GKE_CRED_JSON…"


viccuad:
@prabalsharma, can you have a look? I know it's removing some functionality, but it works nevertheless, and this restores the contract.

prabalsharma:
@viccuad 
GKE kubeconfigs are bit different. They have to be generated using  kube clusterreference file. As each generated kubeconfig file has a time stamp you can not use a generated kubeconfig between job as they might get expired in middle of a run. 

> impossible for the CI to know the name of the cluster you are importing 
you get it from pools. eg: https://github.com/SUSE/cf-ci-pools/blob/gke-kube-hosts/gke-kube-hosts/claimed/ci-gke-290bebf661749b86

> (nor it needs it anyways)
its need to generate kubeconfig on each concourse container

HartS:
Echoing what @prabalsharma has said; at the time we introduced the ClusterReference file as way to get a definitely-valid kubeconfig for a GKE cluster, there was no way to get a kubeconfig for GKE cluster which did not expire (or if there was, it wasn't discoverable online)

Last year, a Google Cloud engineer published [this blog post](https://ahmet.im/blog/authenticating-to-gke-without-gcloud/) which actually describes a way to get this. So if we'd like to move away from a ClusterReference file and instead use configs which don't expire in our pools, that's a possibility.

This would certainly be useful if we might end up running gke operations from an environment where the GKE credentials won't be available, but I'm not convinced we need this work until then; currently, GKE clusters are either accessed via CI (which has the appropriate credentials to get an expiring kubeconfig via the traditional approach with `gcloud container clusters get-credentials`) or locally when using Catapult (in which case gcloud credentials must be provided in order to perform GKE operations anyway).

viccuad:
> its needed to generate kubeconfig on each concourse container

Could we just make it random, and track clusters with tags?

PR #246, opened 2020-07-09
Make an airgap simulation available for CaaSP clusters

HartS:


Makefile
```diff
@@ -171,7 +179,7 @@ kubecf-build: ##@kubecf Build chart from source and install KubeCF
 kubecf-clean: ##@kubecf Only delete installation of KubeCF & related files
 	$(MAKE) -C modules/kubecf clean
 
-.PHONY: sc	f
+.PHONY: scf
```
viccuad, line 182: arg, this should have been

```suggestion
.PHONY: kubecf
```

seems that I introduced it long ago! hopefully it only created some `catapult/{sc,f}` files because of the phony. Sorry.

HartS:
This has some more work pending.. since I don't think I can block it from being merged while I add final changes, will close this for right now

PR #247, opened 2020-07-10
Add a target to check if the stratos ui is reachable

greygoo:


modules/stratos/reachable.sh
```diff
@@ -0,0 +1,55 @@
...
+      ok "Reachable via hostname https://console.${domain}:${port}" 
+      break
+   fi
+   
+   n=$[$n+1]
+   sleep 60
+done
+
+if [ $exit_code -ne 0 ] ; then
+   fail "Not reachable via IP"
```
viccuad, line 54: Here it should name hostname instead of IP:

```suggestion
   fail "Not reachable via hostname https://console.${domain}:${port}"
```

modules/stratos/Makefile
```diff
@@ -20,5 +20,9 @@ install:
 upgrade:
 	./upgrade.sh
 
+.PHONY: reachable
+reachable:
+	./reachable.sh
+
 .PHONY: all
-all: clean chart gen-config install
+all: clean chart gen-config install reachable
```
viccuad, line 28: I would prefer to not add `reachable` here, for those clusters were you can't access the DNS domain outside of the cluster for security reasons (such as the ones in kubecf pipeline).

modules/stratos/reachable.sh
greygoo: that was the idea, make first sure it can be reached at all and if that is true, check if it is reachable via hostname.

viccuad: Yes, it's just that in this spot, we have already checked by IP, and we are checking by domain now, so the printed output should reflect that.

PR #250, opened 2020-07-14
Add annotate function for stratos and use it.

greygoo:


viccuad:
This look great! merging!

PR #251, opened 2020-07-15
feat: backend: custom owner

mook-as:
Allow override of the owner, rather than always forcing `$(whoami)` to be used.  This is useful for CI, where we can more properly identify the cluster rather than whatever user CI happens to run under.  This is also useful for cases where our login name does not match e.g. the GitHub identifier or similar.


prabalsharma:
LGTM

PR #252, opened 2020-07-17
Add stress/benchmark tool from Ettore as catapult target

greygoo:


modules/tests/stress.sh
```diff
@@ -0,0 +1,17 @@
...
+
+[ ! -d "$STRESS_FOLDER" ] && git clone --recurse-submodules "$STRESS_TEST_REPO" "$STRESS_FOLDER"
+
+pushd "$STRESS_FOLDER" || exit
+
+if [ -n "$EKCP_PROXY" ]; then
+    export https_proxy=socks5://127.0.0.1:${KUBEPROXY_PORT}
+fi
+
+bash ./start_benchmark.sh
```
viccuad, line 17: Nitpicking, the bash call shouldn't be needed:

```suggestion
./start_benchmark.sh
```

viccuad, line 17: Nitpicking: I would add at the end an informative  `ok Stress benchmark finished, results in foo and bar` or something similar.

PR #253, opened 2020-07-17
fix: kubecf: Put LB DNS annotations in the config

mook-as:
Add the annotations required to set up DNS names for the load balancers into the helm values file we generate, so that we don't need to patch it afterwards.  However, we can't remove the patching yet as currently the Eirini chart has no way to attach extra annotations; we still need to do something better when Eirini is enabled so we can use ssh correctly.

### DIFF
--- a/include/func.sh
+++ b/include/func.sh
@@ -312,27 +312,6 @@ external_dns_annotate_stratos() {
             "external-dns.alpha.kubernetes.io/hostname=console.${domain}, *.console.${domain}"
 }
 
-external_dns_annotate_kubecf() {
-    local ns=$1;shift
-    local domain=$1;shift
-    # for kubecf
-    kubectl annotate svc router-public \
-            -n "$ns" \
-            "external-dns.alpha.kubernetes.io/hostname=${domain}, *.${domain}"
-    if [[ "${ENABLE_EIRINI}" == true ]] ; then
-        kubectl annotate svc eirinix-ssh-proxy \
-                -n "$ns" \
-                "external-dns.alpha.kubernetes.io/hostname=ssh.${domain}"
-    else
-        kubectl annotate svc ssh-proxy-public \
-            -n "$ns" \
-            "external-dns.alpha.kubernetes.io/hostname=ssh.${domain}"
-    fi
-    kubectl annotate svc tcp-router-public \
-            -n "$ns" \
-            "external-dns.alpha.kubernetes.io/hostname=*.tcp.${domain}, tcp.${domain}"
-}
-
 external_dns_annotate_scf() {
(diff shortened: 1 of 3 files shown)


viccuad:
@mook-as do you know if Eirini chart can attach extra annotations nowadays? It would be nice to have this merged.

mook-as:
@viccuad Oh! Given that the Eirini chart isn't used anymore (we're just using the images directly), this should be totally doable now.  Thanks for the reminder!

mook-as:
The only relevant thing for Eirini is ssh-proxy, and that already reads the correct annotation, so there shouldn't be anything additional here.

viccuad (commented):
For testing on Eirini it needs rebase onto master, as we need `install_stacks: [sle15]` which is already there.

after helm install of kubecf, I get:
```
secret/susecf-scf.var-cf-admin-password created
secret/var-cf-admin-password created
coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>
coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>
coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>
coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>
coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>
coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>
```

The deployment succeeds anyway.

Seems that it comes from https://github.com/helm/helm/issues/4514. Using the current HELM_VERSION from master, which is 3.1.1.

viccuad:
We can drop:
https://github.com/SUSE/catapult/blob/18d4ef3c71c38d1e999d1c8ec7d9f65a5cacad14/modules/kubecf/install.sh#L100-L102

And therefore drop:
https://github.com/SUSE/catapult/blob/18d4ef3c71c38d1e999d1c8ec7d9f65a5cacad14/include/func.sh#L315-L334

viccuad:
I have rebased and applied changes on branch `marky/dns-annotations`, from this fork branch. We could open a PR from that one and close this one, or update this one.

Opened https://github.com/cloudfoundry-incubator/kubecf/pull/1316 in kubecf to fix the warnings.

mook-as:
@viccuad The fork branch has been updated.

FWIW, pushing to the fork is fine too; the PR is set so that people with write access to the original repo can push to the fork.

viccuad:
> FWIW, pushing to the fork is fine too; the PR is set so that people with write access to the original repo can push to the fork.

Agh, I didn't realize. Thanks for looking into it! merging :).

PR #258, opened 2020-07-20
Add catapult targets for simulated airgapped environment, and kubecf/CAP installation changes to allow this

HartS:
Add catapult targets to "airgap" and "reconnect" a CaaSP cluster.
SSH access to CaaSP nodes is required.

This work assumes `DOCKER_REGISTRY` will only be set when testing
kubecf/CAP installation on an "airgapped" CaaSP cluster (one which is
physically cut off from the internet, or has had internet egress
disabled with the `module-experimental-airgap-up` make target)

Since the airgap simulation leverages the namespaced CiliumNetworkPolicy
resource, the restrictive policies become non-functional when the kubecf
deployment is cleaned. Therefore, `module-experimental-airgap-down` has
been added as a prerequisite of the `kubecf-clean` target. Additionally,
all helm parameters will now be set to ensure no images/resources are
pulled from the internet. Feedback is requested from anyone who is
using catapult with the `DOCKER_REGISTRY` setting overridden


Makefile
```diff
@@ -171,7 +179,7 @@ kubecf-build: ##@kubecf Build chart from source and install KubeCF
 kubecf-clean: ##@kubecf Only delete installation of KubeCF & related files
 	$(MAKE) -C modules/kubecf clean
 
-.PHONY: sc	f
+.PHONY: scf
```
viccuad, line 182: This fix should be:

```suggestion
.PHONY: kubecf
```

modules/experimental/Makefile
```diff
@@ -8,3 +8,11 @@ eirini_release:
 .PHONY: eirinifs
 eirinifs:
 	./eirinifs.sh
+
+.PHONY: airgap-up
+airgap-up: airgap-down
```
viccuad, line 13: :100:  calling airgap-down here too

modules/experimental/airgap_down.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+#!/usr/bin/env bash
```
viccuad, line 3: We have a repeated line here

```suggestion
```

modules/experimental/airgap_down.sh
```diff
@@ -0,0 +1,46 @@
...
+
+. ./defaults.sh
+. ../../include/common.sh
+. .envrc
+
+if [[ ${BACKEND} != "caasp4os" ]]; then
```
viccuad, line 9: I would like to use the cap-values `platform` key better, that way it will work if people don't call it with the BACKEND var. See https://github.com/SUSE/catapult/blob/db7da28355dfb2d7034de67bbc50aff36ef396a6/backend/caasp4os/deploy.sh#L120

modules/experimental/airgap_up.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+. ./defaults.sh
+. ../../include/common.sh
+. .envrc
+
+if [[ ${BACKEND} != "caasp4os" ]]; then
```
viccuad, line 7: same here, with https://github.com/SUSE/catapult/blob/db7da28355dfb2d7034de67bbc50aff36ef396a6/backend/caasp4os/deploy.sh#L120

modules/experimental/cilium-block-egress.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: "cilium.io/v2"
```
viccuad, line 1: Maybe this file should go into `catapult/kube/airgap/` to be cleaner, all yaml files are there.

modules/kubecf/install.sh
```diff
@@ -73,11 +86,16 @@ wait_for "kubectl get crd boshdeployments.quarks.cloudfoundry.org -o name"
 info "Test CRDs are ready"
 #wait_for "kubectl apply -f ../kube/cf-operator/boshdeployment.yaml --namespace=scf"
 wait_for "kubectl apply -f ../kube/cf-operator/password.yaml --namespace=scf"
-wait_for "kubectl apply -f ../kube/cf-operator/qstatefulset_tolerations.yaml --namespace=scf"
+if [[ "${DOCKER_REGISTRY}" != "registry.suse.com" ]]; then
```
viccuad, line 89: Shouldn't this be as follows? as we don't want to wait if we change the registry, as it will fail.
```suggestion
if [[ "${DOCKER_REGISTRY}" == "registry.suse.com" ]]; then
```

modules/kubecf/install.sh
```diff
@@ -73,11 +86,16 @@ wait_for "kubectl get crd boshdeployments.quarks.cloudfoundry.org -o name"
...
+  # qstate_tolerations fails when internet connectivity is disabled.
+  wait_for "kubectl apply -f ../kube/cf-operator/qstatefulset_tolerations.yaml --namespace=scf"
+fi
 wait_ns scf
 #wait_for "kubectl delete -f ../kube/cf-operator/boshdeployment.yaml --namespace=scf"
 wait_for "kubectl delete -f ../kube/cf-operator/password.yaml --namespace=scf"
-wait_for "kubectl delete -f ../kube/cf-operator/qstatefulset_tolerations.yaml --namespace=scf"
+if [[ "${DOCKER_REGISTRY}" != "registry.suse.com" ]]; then
```
viccuad, line 96: Same here:
```suggestion
if [[ "${DOCKER_REGISTRY}" == "registry.suse.com" ]]; then
```

modules/kubecf/install.sh
```diff
@@ -54,11 +53,25 @@ else
     operator_install_args+=(--set "global.operator.watchNamespace=scf")
 fi
 
+if [[ "${DOCKER_REGISTRY}" != "registry.suse.com" ]]; then
```
viccuad, line 56: Nitpicking: I would move this block after the `echo "Installing CFO"` line.

modules/kubecf/install.sh
```diff
@@ -90,9 +108,20 @@ kubectl create secret generic -n scf susecf-scf.var-cf-admin-password --from-lit
 ## CF-Operator >= 4 don't have deployment name in front of secrets name anymore
 kubectl create secret generic -n scf var-cf-admin-password --from-literal=password="${CLUSTER_PASSWORD}"
 
+kubecf_install_args=(--values scf-config-values.yaml)
```
viccuad, line 111: This concerns me to be honest. So far, we had made a point on saving all the state that we are going to install in the values.yaml, and this departs from it. I think this is painful, and will bring us headaches in the future, both developing and debugging already deployed clusters, because if we get into the habit of setting deployment options inline, you never know on what state the cluster was deployed.

I think these changes should go in `modules/kubecf/gen-config.sh`. Yes, we lose functionality of airgapping "on the fly", but production deployments don't need to be airgapped "on the fly", you configure the deployment to be airgapped, and then you deploy.

modules/experimental/cilium-block-egress.yaml
HartS: I wanted to keep experimental work in the experimental folder. Even though I think this is fairly stable now, I think more people would need to use it to be promoted to a stable api feature

HartS:
Thanks for the fast turnaround @viccuad . Then I'll check with Prabal to see if he wants to do a quick pass, deal with linting issues, and we can get it merged today

modules/kubecf/install.sh
HartS: I'm unsure about this one. The other args are all set before this is printed, and it should be the last thing that prints before actually the helm chart/release.

HartS:
Addressed PR feedback as well as concerns around how we're handling config customization in general which I discussed with @viccuad directly. The way this is now handled is to apply changes to the config structure by merging "customization" objects as many times as we need to with the jq `*` operator, before finally writing the whole config values back to yaml.

modules/experimental/airgap_down.sh
HartS: `BACKEND` always needs to be set for catapult. That line you linked requires `caasp4os` to be the `BACKEND` in order to generate that configmap. Users don't need to set backend every time they run catapult make targets however, the backend info is in the .envrc  file

HartS: Checking `BACKEND` is more reliable than checking the configmap as well, as you may need to run certain targets when the kubernetes cluster may be unavailable, like kubeconfig, charts, gen-config, cleaning, etc.

PR #263, opened 2020-07-29
remove hardcoded value of SCF_CHART from kubecf install

prabalsharma:


viccuad:
Yes, this line is not needed anymore, and may break things!

It seems that it was there for use in https://github.com/SUSE/catapult/blob/163504afa0767e5c0e8e85b99409642810ce0d2b/modules/kubecf/install.sh#L93

It was needed in the past because a SCF_CHART could come from a repo, or a local folder. By merging https://github.com/SUSE/catapult/pull/259, we now either untar, download and untar, or helm fetch the chart, putting it always on the same spot in `buildfolder`.

viccuad:
Actually, I don't think that line should be removed.

It needs to be hardcoded: `make kubecf-chart` makes sure to either fetch and untar, copy from a folder, or untar a passed tgz from a URL or file __into__ the `buildfoo/kubecf` folder.

`modules/kubecf/install.sh` doesn't know (and can't know) from where the chart it is installing comes from.


What is the use case for this?
I think that there is a misunderstanding, as catapult only guarantees that options are consistent when you call the big states: `kubecf, stratos, metrics, scf, etc`. If you want to install kubecf from a local chart, you need to do either:
`SCF_CHART=abs path to file make kubecf`
or ``SCF_CHART=abs path make kubecf-chart` and then `make kubecf-install`.

prabalsharma:
@viccuad closing this and opened https://github.com/SUSE/catapult/issues/265

PR #264, opened 2020-07-30
Don't exit on name lookup errors while logging in

dragonchaser:
This PR makes sure that the catapult script does not exit on the pipeline when the nslookup for cf-login fails.


jimmykarily:
But does the login succeed?

PR #266, opened 2020-08-03
turn off credhub tests for cats-eirini

jbuns:
This is according to the testing done here:
https://github.com/cloudfoundry-incubator/kubecf/issues/1007#issuecomment-656753374

Setting credhub_mode to `off` should set this:
https://github.com/HCL-Cloud-Native-Labs/cf-acceptance-tests/blob/master/helpers/config/config_struct.go#L844
and this:
https://github.com/HCL-Cloud-Native-Labs/cf-acceptance-tests/blob/master/helpers/config/config_struct.go#L848
to `false` and will skip credhub tests:
https://github.com/HCL-Cloud-Native-Labs/cf-acceptance-tests/blob/master/cats_suite_helpers/cats_suite_helpers.go#L253

Consequently, the `diego` check is unneeded and that PR is here:
https://github.com/SUSE/cf-acceptance-tests/pull/2

### DIFF
--- a/modules/tests/kubecf-test.sh
+++ b/modules/tests/kubecf-test.sh
@@ -138,7 +138,7 @@ create_cats_internetless_secret() {
     kubectl get secret -n "${KUBECF_NAMESPACE}" "${cats_secret_name}" -o json  | \
     jq -r '.data."properties.yaml"' | base64 -d | \
     yq w - "instance_groups.(name==acceptance-tests).jobs.(name==acceptance-tests).properties.acceptance_tests.include" '=internetless' | \
-    yq w - "instance_groups.(name==acceptance-tests).jobs.(name==acceptance-tests).properties.acceptance_tests.credhub_mode" '"off"' | \
+    yq w - "instance_groups.(name==acceptance-tests).jobs.(name==acceptance-tests).properties.acceptance_tests.credhub_mode" 'off' | \
     base64 -w 0 )"
 
   cats_updated="$(echo "$cats_secret" | yq w - 'data[properties.yaml]' $cats_updated_properties)"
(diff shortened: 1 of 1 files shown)


jandubois:
@jbuns I don't think this change is working; it seems to be breaking the tests now that it is included in kubecf CI:

https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/cats-internetless-diego-pr/builds/2

>Failed to unmarshal: json: cannot unmarshal bool into Go struct field config.credhub_mode of type string

According to the docs the only valid settings for `credhub_mode` are `assisted` and `non-assisted` according to https://github.com/cloudfoundry/cf-acceptance-tests#credhub-modes

The choise of the string `off` without quotes is the real issue here, as in YAML 1.1 this will be parsed as the boolean value `false`: https://yaml.org/type/bool.html. This breaks deserialization into a string.

I believe in YAML 1.2 this would be parsed as the string `off`, and things would work, just because `"off" != "assisted"`, but I think it would be better to only use the documented values.

@viovanov can you revert this commit in catapult and update the pipeline to pin to a catapult version without this change, to get CI working again?

viovanov:
@jandubois I think @jbuns is going after the effect described in this message [1].
So I'll use the value `skip-tests` for the mode and see if that helps first.

[1] https://github.com/cloudfoundry/cf-acceptance-tests/blob/5e235612d51f97b76881ac6120c765805ade83ef/helpers/skip_messages/skip_messages.go#L15

jbuns:
@jandubois I just saw the mistake and yes, I should have left the "" on '"off"' so that it gets parsed as a string rather than a boolean. Apologies for that.

@viovanov just saw the fix merged to master, thanks for that.

Issue #265, opened 2020-07-30
Using separate kubecf and cfo charts is not supported


prabalsharma:
looks like catapult don't support individual tgz for kubecf and operator. I guess it only supports bundle (kubecf + cfo)
example:
```
SCF_CHART=/tmp/build/0a06c48c/helm-chart.kubecf-chart/kubecf-2.2.2.tgz
OPERATOR_CHART_URL=/tmp/build/0a06c48c/helm-chart.cfo-chart/cf-operator-4.5.6+0.gffc6f942.tgz

+ cp -rfv /tmp/build/0a06c48c/helm-chart.kubecf-chart/kubecf-2.2.2.tgz chart
'/tmp/build/0a06c48c/helm-chart.kubecf-chart/kubecf-2.2.2.tgz' -> 'chart'
+ echo /tmp/build/0a06c48c/helm-chart.kubecf-chart/kubecf-2.2.2.tgz
+ grep -q tgz
+ '[' -f chart ']'
+ tar -xvf chart -C ./
kubecf/Chart.yaml
.
.
kubecf/charts/eirini-extensions/templates/ssh.yaml
kubecf/charts/eirini-extensions/.helmignore
+ rm -f chart
+ for file in *tgz
+ tar -xvf '*tgz' -C ./
tar: *tgz: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
make[1]: *** [Makefile:9: chart] Error 2
make[1]: Leaving directory '/tmp/build/0a06c48c/catapult/modules/kubecf'
make: *** [Makefile:176: kubecf] Error 2
```

We actually don't need chart.sh in this case. As we can directly do helm install on these tgz files as they are not bundled, rather are actual charts.

viccuad:
I agree with this, and I suspect that losing the possibility of using separate charts may be a regression we introduced in https://github.com/SUSE/catapult/pull/259.

prabalsharma:
fixed

PR #268, opened 2020-08-04
Fix kubecf-chart target when passed a local file

viccuad:
Don't try to untar second tgz files if there is none (that would be the case of
kubecf bundles).

In the case of passing for example:
  SCF_CHART=/tmp/build/0a06c48c/helm-chart.kubecf-chart/kubecf-2.2.2.tgz
it already happens in a previous check.


viccuad:
Closes https://github.com/SUSE/catapult/issues/265

PR #273, opened 2020-08-14
Changes for improved terraform usage

HartS:
This PR contains several noteworthy improvements to how catapult uses terraform to deploy aks/gke/eks:

- tfvars files are now created in a shared `tfsetup` target, which is a dependency of `deploy` and `clean` targets for aks, gke, and eks (but the tfsetup target is different for each backend)
- when the `TF_KEY` variable is set, a `backend.tf` file is created as part of the `tfsetup` target, which tells terraform to store state in s3. `terraform init` is run on both deploy and clean targets to initialize the terraform directory. This means for clean, when `TF_KEY` is set, the state will be fetched from s3 before  `terraform destroy` is called.
- The kube version to install has been bumped for AKS from `1.18.2` to `1.18.4` due to moving AKS version support
- The terraform version used has been upgraded from `0.12.9` to `0.12.29` to match what CI is using
- There is no more manual handling of tfstate from the clean steps. tfstate will either exist from running the deploy locally, or will be fetched when using an s3 backend, when `TF_KEY` is set
- The branch of https://github.com/suse/cap-terraform to be used has been parameterized with the `CAP_TERRAFORM_BRANCH` global variable. The default branch to use is `cap-ci`, but exporting this to something else will allow testing catapult integration with changes in cap-terraform which have yet to be merged.
- Some error case output has improved. Catapult will also no longer try to run `aws configure` (which involves manual interaction)
- `terraform apply` will be re-run if it encounters a failure on the first try, to address situations where the first `apply` failed due to resources not being ready when expected.

Additionally, a new `module-experimental-tf-force-clean` target has been added to facilitate cleanup of clusters which may have failed in CI. For example, if concourse fails to add the cluster information to the aks/gke/eks backends pool, the build will fail and won't be able to clean up the cluster either. This target expects `CLUSTER_NAME` to be set, as well as `BACKEND` if it can't be inferred from the `CLUSTER_NAME`, and must be able to access a kubernetes target containing secrets for the concourse deployment, mirroring those referenced from the related PR to https://github.com/suse/cap (in other words, a concourse target deployed to kubernetes which feeds secrets to the pipelines via secrets in the `concourse-main` namespace). For AKS and GKE backends, a `KUBECFG` variable (for consistency with the `kubeconfig` target) needs to be set to a valid kubeconfig which catapult/terraform can use to access the kube api.

When run, the `module-experimental-tf-force-clean` target will generate the tfvars files for the cluster which needs to be force cleaned, create a `backend.tf` file which tells terraform to read the state from an s3 backend, and run `terraform destroy`. The name of the terraform state file in s3 needs to match the cluster name


modules/experimental/tf_auto_deploy.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+# Deploys an eks/aks/gke cluster with values taken from concourse secrets. For local use, but deploys a k8s cluster like CI
+
+if [[ -n "${CLUSTER_NAME}" ]]; then
+    echo "module-experimental-tf-force-clean requires CLUSTER_NAME to be unset"
```
prabalsharma, line 6: err. should be deploy

PR #274, opened 2020-08-19
Parameterize helm repo installation versions

HartS:
When installing from the suse helm repo with `SCF_CHART=from_repo`, also
support setting KUBECF_VERSION and CF_OPERATOR_VERSION to specify what
version of the charts to install. Useful for testing upgrades


prabalsharma:
LGTM

PR #277, opened 2020-09-03
Don't use CLUSTER_PASSWORD in kubecf, it's a security risk

viccuad:
Drop setting var-cf-admin-password manually, let the deployment create the
secret.
Then, when needed, just read the secret with kubectl.


viccuad:
Tested successfully on a GKE cluster.

PR #281, opened 2020-09-15
feat: AKS: allow explicit configuration of ssh key

mook-as:
This allows an explicit SSH key to be used, rather than whatever is the first in the ssh agent.

This is useful if the first key in the keychain isn't accepted by Azure (e.g. I don't think they like ed25519 yet).

The existing behaviour is unchanged.


viccuad (commented):
LGTM. I would like to add
```
AZURE_SSH_KEY="${AZURE_SSH_KEY:-}" # optional ssh key file for azure. If not set, take first in keyring.
```
 in `backend/aks/defaults.sh` to keep track of it.

mook-as:
Thanks for the review! That sounds totally sane, so I've added a commit to address that.

viccuad (approved):

PR #283, opened 2020-09-15
feat: AKS: remove resource group on clean

mook-as:
We create a resource group in AKS on deployment; for mirroring, we should also remove it when cleaning up.

### DIFF
--- a/backend/aks/clean.sh
+++ b/backend/aks/clean.sh
@@ -22,6 +22,9 @@ if [ -d "$BUILD_DIR" ]; then
     terraform init
     terraform destroy -auto-approve
     popd || exit
+    if az group show --name "${AZURE_RESOURCE_GROUP}" 2>/dev/null; then
+        az group delete --name "${AZURE_RESOURCE_GROUP}" --yes
+    fi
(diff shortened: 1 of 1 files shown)


viccuad (changes requested):
I'm not sure we need to remove the resource group.

From cap-terraform's readme, the resource group is a precondition for terraform, and it is personal:
https://github.com/SUSE/cap-terraform/blob/8831476c6c81758c54c162d780c5fcf039514b58/aks/README.md

> This resource group will contain all resources of, and in support of your SUSE CAP cluster. Make note of the name and location which must be supplied in terraform variables.

And it seems that terraform doesn't create it either (following var https://github.com/SUSE/cap-terraform/blob/8831476c6c81758c54c162d780c5fcf039514b58/aks/variables.tf.json#L19).

So I see it as auth setup for the user, and in that case I think we should let the user decide when and how the resource group will be deleted, as it can be reused between deployments.

mook-as:
The problem was that, if we _don't_ remove the resource group, the next deployment fails for me (because it already exists).

I'll try again later to get the exact error.

PR #285, opened 2020-09-24
fix cfo upgrade command for CAP 2.1.0

prabalsharma:


satadruroy (approved):
LGTM.

viccuad:
@satadruroy  @prabalsharma  Could you expand what gets fixed with this? It is missing in the commit or PR messages.

The change from `--reuse-values` to `--set "global.singleNamespace.name=scf"` should have no effect, as that was already the case.

And adding 3 times `helm list -A kubectl get pods -A` only adds more cruft in my opinion. If we need to wait for something, we would be better served by adding a sleep with its corresponding comment.

satadruroy:
@viccuad `--reuse-values` for cfo upgrade fails because the templates are structurally different between the operator version corresponding to `2.0.1` and the current version. Besides, `global.singleNamespace.name=scf` is what is used for current operator version whereas what was being used in previous published version was `global.operator.watchNamespace=kubecf`

The rest of the comments seem valid - we can perhaps take care of them in a separate PR but I'll let @prabalsharma comment.

viccuad:
@satadruroy thanks! I didn't recall the config differences between 2.0.1 and current. On top of that, it seems that `--reuse-values` don't get checked by the schema of the chart version that was deployed and matches the values, but the schema of the chart you are upgrading, which could be dangerous. And maybe we shouldn't use it if we are upgrading between different versions of the chart (see https://cloudfoundry.slack.com/archives/CQ2U3L6DC/p1601314809489600),

viccuad:
simplified output of the upgrade target in https://github.com/SUSE/catapult/pull/288.

PR #286, opened 2020-09-25
fix: Reduce cf-o wait sleep from 240 to 10

viccuad:
Change introduced in:
https://github.com/SUSE/catapult/commit/fc8e007992bafd9f06dd37638e92e4128de0f938


viccuad:
Tested successfully in kind and CaaSP.

dragonchaser (approved):
LGTM

PR #288, opened 2020-09-29
fix: Simplify upgrade target output, only show helm ls -A

viccuad:


viccuad:
merging this to consume it on kubecf CI, as we need to bump there to consume https://github.com/SUSE/catapult/pull/285

PR #290, opened 2020-09-29
Attempt to increase timeouts to avoid CI flakiness

viovanov:


jimmykarily (approved):

viccuad:
Closing this; timeouts of cats should be changed in cats config created in https://github.com/SUSE/catapult/blob/master/modules/tests/kubecf-test.sh. (this PR changes running cats directly from source inside a pod).

PR #291, opened 2020-09-29
revert: Add vscode default test task to run `make catapult-test`

f0rmiga:
Reverts SUSE/catapult#111.

IDE/editor specifics should never be checked into VCS.


viccuad:
Fair enough. Added long ago in hopes that people would run tests before committing, I think nobody is using it, including me.

Reverting!

f0rmiga:
@viccuad I see. Thanks for reverting. We can always encourage people to set up pre-commit hooks - hard to get everyone to do, though. This is why CI is here! :)
https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

PR #294, opened 2020-10-05
Add default `services` key on cap-values when creating a cluster.

viccuad:
Set a `services` key on cap-values when creating a cluster.

It informs on the type of service termination the cluster has:
- lb.
- ingress.
- hardcoded: use the public ip of the nodes.

This gets used in `modules/kubecf/gen-config.sh` to construct the correct `values.yaml`.

One can overwrite it by setting `KUBECF_SERVICES`, see `modules/kubecf/defaults.sh`.


svollath:
LGTM

svollath (approved):

Issue #295, opened 2020-10-05
Catapult in not setting dns annotations in CAP pipelines


prabalsharma:
from CI: post deployment of kubecf
```
Request error: Get https://api.prabal-gke.ci.kubecf.charmedquarks.me: dial tcp: lookup api.prabal-gke.ci.kubecf.charmedquarks.me on 172.31.2.5:53: no such host
TIP: If you are behind a firewall and require an HTTP proxy, verify the https_proxy environment variable is correctly set. Else, check your network connection.
```

```
286ca24c-1021-41ec-64c0-a3bddf9b24a5:/tmp/build/0a06c48c # cat catapult/buildprabal-gke/scf-config-values.yaml
---
system_domain: prabal-gke.ci.kubecf.charmedquarks.me
install_stacks:
- sle15
- cflinuxfs3
features:
  eirini:
    enabled: false
  autoscaler:
    enabled: false
high_availability: false
testing:
  brain_tests:
    enabled: true
  cf_acceptance_tests:
    enabled: true
  smoke_tests:
    enabled: true
  sync_integration_tests:
    enabled: true
properties:
  acceptance-tests:
    acceptance-tests:
      acceptance_tests:
        timeout_scale: 3
        ginkgo:
          slow_spec_threshold: 300
          extra_flags: ''
          nodes: 3
          flake_attempts: 5
  brain-tests:
    acceptance-tests-brain:
      acceptance_tests_brain:
        verbose: 'false'
        in_order: 'false'
        include: ''
        exclude: ''
```

No dns annotation https://github.com/SUSE/catapult/blob/master/modules/kubecf/gen_config.sh#L44-L67

```
286ca24c-1021-41ec-64c0-a3bddf9b24a5:/tmp/build/0a06c48c # source catapult/buildprabal-gke/.envrc 
286ca24c-1021-41ec-64c0-a3bddf9b24a5:/tmp/build/0a06c48c # kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["services"]'
null
286ca24c-1021-41ec-64c0-a3bddf9b24a5:/tmp/build/0a06c48c # 
```

prabalsharma:
proper service entry did get created but then somehow don't exist! 

```
ℹ  🚀  gke ☸ prabal-gke  🎂  ./deploy.sh ➤  Configuring deployed GKE cluster…
configmap/cap-values created
{
    "apiVersion": "v1",
    "data": {
[...]

PR #297, opened 2020-10-06
Don't delete `services` on cap-values anymore

viccuad:
"services", as "domain" in the cap-values, are part of the cluster environment.
Not something you can easily change on a whim when deploying kubecf.

If one wants to change it, one needs a different cluster creation.

Fixes https://github.com/SUSE/catapult/issues/295, introduced with https://github.com/SUSE/catapult/pull/294.


svollath (approved):
LGTM

PR #298, opened 2020-10-06
Add ingress cert and key to values yaml

viccuad:
Expect `ingress-cert` and `ingress-key` to contain what the cert and key in `cap-values` configmap.

If `services: ingress` in `cap-values` configmap (or alternatively, `KUBECF_SERVICES=ingress` as override), read the cert and key from `cap-values` and put it in the values yaml.

This allows us to deploy kubecf against k8s with ingress, provided that they have in cap-values:
```
data:
  services: ingress
  ingress-cert: |
    foo
  ingress-cert-key: |
    bar
```

### DIFF
--- a/modules/kubecf/gen_config.sh
+++ b/modules/kubecf/gen_config.sh
@@ -15,19 +15,6 @@ else
 fi
 domain=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["domain"]')
 
-if [ "${services}" == "ingress" ]; then
-    INGRESS_BLOCK="ingress:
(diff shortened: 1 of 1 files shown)


viccuad:
Tested by:
`KUBECF_SERVICES=ingress make kubecf-gen-config` -> cert and key appear
`KUBECF_SERVICES=lb make kubecf-gen-config` -> no ingress block, good
same if editting `services: lb` (or ingress) in cap-values configmap.

svollath (approved):
LGTM

PR #299, opened 2020-10-07
Also set services.X.externalIPs when `hardcoded`

viccuad:
We need services.X.externalIPs set for hardcoded backends, such as kind.


dragonchaser (approved):
lgtm

PR #300, opened 2020-10-08
Clean all helm releases in minibroker ns

viccuad:
MITS may leave releases (I'm not sure if testsuite is expected to clean them on
failure).

See: https://github.com/SUSE/minibroker-integration-tests/issues/24


svollath (approved):
lgtm

PR #301, opened 2020-10-08
Fix minikube deployments

viccuad:
Change MINIKUBE_VERSION to v1.9.2.
Use same minikube options and iso as scripts/kubecf.
Don't set rolebindings or storageclass, not needed.
Set `services=hardcoded` on minikube, to use the public ip as domain.


viccuad:
Tested by deploying on minikube, then deleting the cluster.

Merging because we need it for airgap testing, and it doesn't touch anything besides fixing minikube.

PR #303, opened 2020-10-09
Add workaround for upgrade of cap-2.1.0-rc2

viccuad:
Add a folder /modules/kubecf/workarounds/ that will contain all workarounds to
be applied.

Apply workarounds/pre-upgrade-cap-2.1.0-rc2.sh before upgrade process.


viccuad:
This makes upgrades work on cap CI. Will have no changes on KubeCF CI, as there the catapult commit is pinned.

mudler (approved):

PR #305, opened 2020-10-09
Fix tls key name and suppress trailing newline

svollath:
Expects the following key names in configmap, for ingress tls:

```
data:
  ingress-cert: |
    -----BEGIN CERTIFICATE-----
    ...
    -----END CERTIFICATE-----
  ingress-key: |
    -----BEGIN RSA PRIVATE KEY-----
    ...
    -----END RSA PRIVATE KEY-----
  services: ingress

When those are read from configmap, `jq -j` will enter rawmode, but suppress a trailing newline in the output


viccuad (approved):
LGTM

PR #306, opened 2020-10-09
Add wait-autoscaler-dns-clbo-cap-2.1.0-rc2.s workaround

viccuad:
Adds `wait-autoscaler-dns-clbo-cap-2.1.0-rc2.sh` workaround in wait_ns that:
if in, autoscaler, and any pods starting with `asactors` or `asapi` are in CLBO, then delete them.

Prettify workaround `pre-upgrade-cap-2.1.0-rc2.sh`.


viccuad:
The workaround works perfectly fine when it's applied, but the pods keep getting into CLBO in my test minikube. So I wonder of the usage of merging this.

viccuad:
Closing, not needed anymore.

PR #309, opened 2020-10-15
Add deletion of cf-operator webhook to kubecf-clean target

viccuad:


svollath:
According to https://github.com/cloudfoundry-incubator/quarks-operator/issues/1196 we do need to clean `cf-operator-hook-<cf-operator-namespace>` for:

`validatingwebhookconfigurations.admissionregistration.k8s.io`, and
`mutatingwebhookconfigurations.admissionregistration.k8s.io`

Guess it would also be fine to delete any webhook for both, since multiple operators aren't supported anyhow ...

viccuad:
Thanks!
I've added a deletion for the mutatingwebhookconfigurations.admissionregistration.k8s.io too.

svollath (approved):
LGTM

PR #279, opened 2020-09-11
Bump helm to 3.3.3

viccuad:
Bump HELM_VERSION global default to 3.3.1, plus some minor cleanups.

PR #311, opened 2020-10-23
Revert "Bump helm to 3.3.3"

viccuad:
Reverts SUSE/catapult#279.

We have found problems both manually and in the pipeline jobs, with CaaSP, that point to Helm failing to rend manifests, with Helm 3.3.3 & 3.3.4.


svollath (approved):
LGTM

PR #312, opened 2020-10-26
feat: GKE: Allow adding extra labels

mook-as:
Being able to add extra labels would be useful for CI (in particular, making it easier for CI to find its builds and which build maps to which cluster).

This is in support of https://github.com/cloudfoundry-incubator/kubecf/issues/1144 to get GitHub Actions to work better for KubeCF.

### DIFF
--- a/backend/gke/tfsetup.sh
+++ b/backend/gke/tfsetup.sh
@@ -28,6 +28,7 @@ gcp_dns_sa_key = "$GKE_DNSCRED_JSON"
 cluster_labels = {
     catapult-clustername = "$GKE_CLUSTER_NAME",
     owner = "${OWNER}"
+    ${EXTRA_LABELS}
 }
 cluster_name   = "$GKE_CLUSTER_NAME"
 k8s_version    = "latest"
--- a/backend/gke/tfsetup.sh
+++ b/backend/gke/tfsetup.sh
@@ -38,12 +38,14 @@ HEREDOC
 
 if [ -n "${EXTRA_LABELS}" ] ; then
     jq --raw-output --argjson labels "${EXTRA_LABELS}" '.cluster_labels *= $labels' terraform.tfvars.json \
-        | sponge terraform.tfvars.json
+        > terraform.tfvars.temp.json
+    mv terraform.tfvars.temp.json terraform.tfvars.json
 fi
 
 if [ -n "${GKE_INSTANCE_TYPE}" ] ; then
     jq --argjson type "${GKE_INSTANCE_TYPE}" '.instance_type = $type' terraform.tfvars.json \
-        | sponge terraform.tfvars.json
+        > terraform.tfvars.temp.json
+    mv terraform.tfvars.temp.json terraform.tfvars.json
 fi
 
 if [ -n "${TF_KEY}" ] ; then
--- a/backend/gke/tfsetup.sh
+++ b/backend/gke/tfsetup.sh
@@ -17,31 +17,33 @@ git checkout "${CAP_TERRAFORM_BRANCH}"
 git pull
 
 # Clear out any existing variables file
-echo "" > terraform.tfvars.json
-yq merge --inplace --tojson --prettyPrint terraform.tfvars.json - <<HEREDOC
-project        : "$GKE_PROJECT"
-location       : "$GKE_LOCATION"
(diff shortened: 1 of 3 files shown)


backend/gke/tfsetup.sh
```diff
@@ -28,6 +28,7 @@ gcp_dns_sa_key = "$GKE_DNSCRED_JSON"
 cluster_labels = {
     catapult-clustername = "$GKE_CLUSTER_NAME",
     owner = "${OWNER}"
```
viccuad, line 30: I suppose we need to be more intelligent here when terminating the owner object with a comma, when EXTRA_LABELS is set up.
Maybe it pays off to incrementally build a json with jq or similar.

viccuad (changes requested):

backend/gke/tfsetup.sh
mook-as: Addressed in 7127ab4176f026be59507c2d6d2f4d7e5a81568c — deployed a cluster and got the extra label.  Did this via `yq`, since we already had it as a dependency.

viccuad (approved):

viccuad:
Travis failed with 3 failures, all the same. It seems that since we are setting `DOWNLOAD_CATAPULT_DEPS == false`, it's missing yq. E.g: https://travis-ci.com/github/SUSE/catapult/builds/194015687#L619-L655

yq et all get pulled when calling common-deps in catapult/Makefile. This happens when calling the base cluster targets (`k8s`,`kubeconfig`,`recover`).

This showcases how the current approach for pulling deps and the multi-tier Makefile just complicates things at this point.

Kludgy ideas:
- add yq to the travis image as a kludge (unit-tests *may* be ok as they test by doing `which foo`, and expecting them to be in `buildbar/bin/` if `DOWNLOAD_CATAPULT_DEPS == true`).
- add a yq mock that returns 0 in https://github.com/SUSE/catapult/tree/master/tests/mocks

mook-as:
I worked around Travis with an _even more_ kludgy idea of just using `jq` instead of `yq`.  This is combined with `sponge` as `jq` doesn't have an `--in-place` option, as mentioned [on their bug tracker](https://github.com/stedolan/jq/issues/2066#issuecomment-598253539).

We should probably spawn a task into looking into fixing the dependency system; I started working on it, but got stuck on needing to download `jq`, but then `jq` is needed to [load the configs](https://github.com/SUSE/catapult/blob/15fb17f4fe2a4a388add684204ec8cb6f36cc690/include/func.sh#L174) to [find the build dir](https://github.com/SUSE/catapult/blob/15fb17f4fe2a4a388add684204ec8cb6f36cc690/include/common.sh#L23).  And we need the build dir to know where to download `jq`.

backend/gke/tfsetup.sh
```diff
@@ -17,31 +17,33 @@ git checkout "${CAP_TERRAFORM_BRANCH}"
...
-        - <<< "cluster_labels: ${EXTRA_LABELS}"
+    jq --raw-output --argjson labels "${EXTRA_LABELS}" '.cluster_labels *= $labels' terraform.tfvars.json \
+        | sponge terraform.tfvars.json
 fi
 
 if [ -n "${GKE_INSTANCE_TYPE}" ] ; then
-    yq write --inplace --tojson --prettyPrint terraform.tfvars.json \
-        instance_type "${GKE_INSTANCE_TYPE}"
+    jq --argjson type "${GKE_INSTANCE_TYPE}" '.instance_type = $type' terraform.tfvars.json \
+        | sponge terraform.tfvars.json
```
viccuad, line 46: Perhaps we could we use a tmp file instead of adding a dependency on `moreutils`?

viccuad (approved):
I see what you did there :D. Fair enough, if `jq` is needed early in the process and always expected.

Totally agree with the dependency system. Opened https://github.com/SUSE/catapult/issues/314 for now, feel free to edit it.

Approving, feel free to merge, but I have a little nitpick if time allows: drop sponge by using a temp file.
If not dropping sponge, please update https://github.com/SUSE/catapult/wiki/Local-requirements and add `moreutils` as dependency.

backend/gke/tfsetup.sh
mook-as: Changed to a temporary file in 12c662930c67fd54d48a87b404f0b4f572d5ab99

Issue #310, opened 2020-10-22
Have a target to clean up orphaned clusters.


mook-as:
Context: KubeCF CI via GitHub Actions, https://github.com/cloudfoundry-incubator/kubecf/pull/1502

I would like a catapult target to be able to delete excess clusters.  This would mean:

- Filter by a owner (tag on the cluster)
- Not using the terraform context
- Probably split into two separate targets:
  - One to list all clusters for an owner, along with all their other tags (so that the pipeline could filter out still relevant clusters)
  - One to delete specific clusters

mook-as:
Fixed by #313 — forgot to close.

PR #313, opened 2020-10-26
feat: GKE: Add targets to find & delete resources

mook-as:
Adds two make targets to help CI manage resources:
- `make find-resources`: Finds clusters belonging to a given owner, and also IP addresses and disks that appear to be orphaned.  Outputs a JSON file containing the resources and their labels.
- `make force-clean-cluster`: Given the resource list from `make find-resources` as an input, plus optional additional filters, this will delete them.

The expected usage is:
- CI will call `make find-resources`, taking a snapshot of all the resources that may be unnecessary.
- CI will then list all active runs.  The ordering is intended to ensure we do not delete clusters that happens to have started between these two steps.
- CI will call `make force-clean-cluster`, passing in filters that will additionally exclude active runs.  This ensures that only orphaned resources will be removed.
- CI can also use the output of `make find-resources` to determine if we have enough capacity to start new runs.

This implements #310 in support of https://github.com/cloudfoundry-incubator/kubecf/issues/1144 to make GitHub Actions work better for KubeCF; the overall plan is documented in [kubecf@marky/actions:/.github/workflows/README.md](https://github.com/cloudfoundry-incubator/kubecf/blob/marky/actions/.github/workflows/README.md) as part of https://github.com/cloudfoundry-incubator/kubecf/pull/1502.


backend/gke/find-resources.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# OPTIONS:
+# GKE_CRED_JSON gcloud credentials
+# GKE_PROJECT   gcloud project
+# OWNER         The owner of the clusters to filter on
+# OUTPUT_FILE   File to output to; outputs to standard out if not set.
+
+. ./defaults.sh
+. ../../include/common.sh
+# Do not require .envrc
```
viccuad, line 15: I would require .envrc nevertheless. The `buildfolder/.config/gcloud/* folder` is always setup by https://github.com/SUSE/catapult/blob/master/include/buildir.sh#L37, which means that we can use it, and people always expect catapult to not mess their local settings and installations.

backend/gke/force-clean-cluster.sh
```diff
@@ -0,0 +1,58 @@
...
+# GKE_PROJECT       gcloud project.
+# RESOURCE_LIST     Resources to remove, as generated by `make find-resources`.
+# CLUSTER_SELECTOR  JQ expression for filtering clusters, defaults to all.
+# ADDRESS_SELECTOR  JQ expression for filtering addresses, defaults to all.
+# DISK_SELECTOR     JQ expression for filtering disks, defaluts to all.
+# ACTUALLY_DELETE   Delete will only be executed if this is set to `true`.
+
+. ./defaults.sh
+. ../../include/common.sh
+# Do not require .envrc
```
viccuad, line 19: Same as before.

viccuad (changes requested):
This is cool, I really look forward to it!

Silly nitpick: add no-ops to the backend's Makefiles that don't have these targets implemented, to work around the quirks of the current layout and provide a "not implemented yet" message.

backend/gke/find-resources.sh
mook-as: As far as I know, we don't include `buildir.sh`? In fact, for this there's no reason to have a builddir at all (because we don't need any of the state).

viccuad: `buildir.sh` gets run on `buildfoo` folder creation, when creating the cluster. That may mean, if `DOWNLOAD_CATAPULT_DEPS == true` (the default now), that gcloud gets downloaded and configured inside of the build folder.

Every target until now operates inside `buildfoo`, sourcing `.envrc`. The idea of this is to have a reproducible deployment and testing environment between machines & CI. And it also helps in not wiping out logins our touching config things out of catapult buildfolders.

This is useful when manually QAing things, as one may have 3-5 buildfolder clusters in parallel, with different providers, or same provider and different project because of quota issues.

mook-as: Yes, but the find & delete resources targets don't create new clusters (and in fact it wouldn't make sense to: the point is to remove stale clusters that catapult / terraform lost track of).

It is perfectly reasonable to want to run these targets locally, to look for clusters that may be orphaned; this is why the target by default just prints out the things it will delete.  It's only by opting in to actually deleting things that it can be dangerous.

viccuad: But even if you aren't deleting clusters, you will have revoked and re-logged in with other creds outside of catapult:
https://github.com/SUSE/catapult/pull/313/files#diff-cb76b3114cb1673b6ae4c55c3d7444775c8bd4ca1e72ef80cec7ddc36bf4cabdR19-R24

Also, one idea for the future is to converge all the headers
```
. ./defaults.sh
. ../../include/common.sh
. .envrc
```
from all scripts in `modules/*`, `backend/*` into [include/common.sh](https://github.com/SUSE/catapult/blob/master/include/common.sh), which is the last point in the execution flow before each distinct target.

That way all testing/operations scripts always execute against an already setup test environment, known to be good. This helps further rebases for a cleaned up catapult v2.

mook-as: Okay, I think I will agree to loading the `.envrc`, mainly because we still need a context (a build dir) in which to dump tools we need to download.  While we currently tie this to a cluster, we don't need to actually _build_ the cluster to run the new targets.  And your point about revoking the login is very important too (I've just been ignoring that while testing this, but in retrospect that wasn't the greatest).

mook-as: This should be resolved in fded88a.

viccuad: Thanks! didn't think about "tying the target to an unexistent cluster", that's a good point. But I think the trade-off pays off for now.

Approving.

viccuad (approved):

PR #315, opened 2020-10-29
feat: Add CLUSTER_SERVICES=ingress for minikube ingress deployments

viccuad:
`BACKEND=minikube CLUSTER_SERVICES=ingress make k8s` will get you a minikube with ingress. This sets the correct key `services=ingress` in cap-values configmap, which means that deployments of kubecf et al will get configured for ingress.

not setting `CLUSTER_SERVICES`, or setting `CLUSTER_SERVICES=hardcoded` will get you the previous behaviour, where one uses externalIPs in LB.


dragonchaser (approved):
LGTM

PR #317, opened 2020-11-05
Add all needed targets to perform airgap installs on GKE

viccuad:
This PR enables setting up airgapped clusters on GKE for testing. It contains
targets for setting up a local-registry, pushing the images to the registry, and
airgapping the cluster via k8s networkPolicies.

Their implementation has been consciously made to be BACKEND-independent.

Prerequisites for full airgap installations:
- k8s networking plugin installed (for now, only GKE cluster).
- insecure container registries in k8s internal network accepted (for now, only
  GKE cluster).

Included in this PR:

- Add BACKEND-side targets `airgap-up` and `airgap-down` for GKE and stubs for
  the rest of backends. Those targets can be expanded in the future when each
  backend satisfies the prerequisites.

- Add networkPolicies for isolating the cluster by only allowing network
  connections in the internal network, and external connections on DNS ports,
  for anything on the `default`, `scf`, and `cf-operator` namespaces.
  This network policies are backend-independent, but they need a k8s Network
  plugin present. At the moment, only GKE gets deployed with such, in that case,
  Calico.

- Add a `module-experimental-local-registry` target. This target deploys a
  Docker v2 registry as a service in ns `default`, only exposing it internally
  to the cluster via the ClusterIP and the default
  local-registry.default.svc.cluster.local DNS name. For Kubelet to correctly
  pull images from the local registry, one needs to target its ClusterIP.
  
- Add a `module-experimental-push-imagelist` target. This target copies all
  images listed in any imagelist.txt file in the buildfolder into the local
  registry. This is done by spawning 1 k8s job per image to copy, and waiting
  for all jobs to finish successfully.
  We are using dragonchaser/opensuse-skopeo:latest image, which contains
  podman (with vfs for unpriviliged containers), and skopeo.

- Add a `module-experimental-podman-imagelist` target. This target spawns a pod
  with podman configured with vfs for unprivileged containers, and the
  imagelist.txt files and a testing script inside the /tmp/ folder. Only useful
  for manual testing and developing the airgap.

### DIFF
--- a/modules/experimental/Makefile
+++ b/modules/experimental/Makefile
@@ -22,7 +22,7 @@ podman-imagelist:
 	./podman-imagelist.sh
 
 .PHONY: airgap-up
-airgap-up: airgap-down
+airgap-up:
 	./airgap_up.sh
 
 PHONY: airgap-down
--- a/modules/experimental/Makefile
+++ b/modules/experimental/Makefile
@@ -9,6 +9,10 @@ eirini_release:
 eirinifs:
 	./eirinifs.sh
 
+.PHONY: local-registry
+local-registry:
+	./local-registry.sh
+
 .PHONY: airgap-up
 airgap-up: airgap-down
 	./airgap_up.sh
--- a/modules/experimental/Makefile
+++ b/modules/experimental/Makefile
@@ -13,6 +13,10 @@ eirinifs:
 local-registry:
 	./local-registry.sh
 
+.PHONY: push-imagelist
+push-imagelist:
+	./push-imagelist.sh
+
 .PHONY: airgap-up
 airgap-up: airgap-down
 	./airgap_up.sh
--- a/modules/experimental/Makefile
+++ b/modules/experimental/Makefile
@@ -17,6 +17,10 @@ local-registry:
 push-imagelist:
 	./push-imagelist.sh
 
+.PHONY: podman-imagelist
+podman-imagelist:
+	./podman-imagelist.sh
+
 .PHONY: airgap-up
 airgap-up: airgap-down
 	./airgap_up.sh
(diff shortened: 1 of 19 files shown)


modules/experimental/push-imagelist.sh
```diff
@@ -0,0 +1,62 @@
...
+kind: Job
+metadata:
+  name: ${JOB_NAME}
+  namespace: push-imagelist
+spec:
+  template:
+    spec:
+      containers:
+      - name: copy-image
+        image: dragonchaser/opensuse-skopeo:latest
```
mook-as, line 32: This image doesn't have the labels to indicate where the source comes from; we should probably add that.  I assume that's `modules/experimental/docker/Dockerfile.skopeo` though?

modules/experimental/Makefile
```diff
@@ -9,8 +9,20 @@ eirini_release:
...
+local-registry:
+	./local-registry.sh
+
+.PHONY: push-imagelist
+push-imagelist:
+	./push-imagelist.sh
+
+.PHONY: podman-imagelist
+podman-imagelist:
+	./podman-imagelist.sh
```
mook-as, line 22: Hmm, it's unclear to me why we have two versions.

modules/experimental/airgap_up.sh
```diff
@@ -30,6 +32,13 @@ sudo -s << 'EOS'
   iptables -A OUTPUT -j ACCEPT -d 192.168.0.0/16
   iptables -A OUTPUT -j DROP -d 0.0.0.0/0
 EOS
+EOF
+
+  # Include insecure registry
+  # shellcheck disable=SC2087
+  ssh -T sles@${kube_node} << EOF
+sudo sed 's/^\(CRIO_OPTIONS\s*=\s*\).*$/\1"--insecure-registry=${DOCKER_REGISTRY}"/' \
```
mook-as, line 40: We're wiping any existing options?

mook-as (changes requested):

PR #318, opened 2020-11-06
changes helm-inspect for cfo charts from appversion to version

prabalsharma:
use `version` instead of `appversion` to check cfo version

Test: https://concourse.suse.dev/teams/main/pipelines/prabal-gke-upgrades/jobs/pre-upgrade-deploy-kubecf-diego-gke-sa/builds/11


viccuad:
Thanks! Merging.

PR #320, opened 2020-11-20
cf-operator has been renamed to quarks-operator in qarks v7

jandubois:
Untested; please review carefully! :)

### DIFF
--- a/modules/kubecf/upgrade.sh
+++ b/modules/kubecf/upgrade.sh
@@ -8,7 +8,7 @@ info "Upgrading CFO…"
 helm list -A
 
 OPERATOR_DIR=cf-operator
-if [ -d quarks ]; then
+if [ -f quarks/Chart.yaml ]; then
     OPERATOR_DIR=quarks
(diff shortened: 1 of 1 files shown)


modules/kubecf/upgrade.sh
```diff
@@ -7,7 +7,11 @@
 info "Upgrading CFO…"
 helm list -A
 
-helm_upgrade cf-operator cf-operator/ \
+OPERATOR_DIR=cf-operator
+if [ -d quarks ]; then
```
mook-as, line 11: Do we need to check that it's the right shape?
```suggestion
if [ -f quarks/Chart.yaml ]; then
```

mook-as (approved):

modules/kubecf/upgrade.sh
jandubois: > Do we need to check that it's the right shape?

I don't think it makes any difference, so I'll just merge in your suggested change :)

PR #325, opened 2020-11-26
feat: backend: GKE: Remove stale DNS entries too

mook-as:
This adds the ability for `make find-resources` and `make force-clean-cluster` to locate stale DNS records for cluster that no longer exist, and to delete them.

This assumes that the clusters are named `kubecf-ci-<owner>` where `<owner>` contains solely of letters and numbers.  See https://github.com/cloudfoundry-incubator/kubecf/pull/1630 for the corresponding change.

To be honest, this is a bit too much use of `jq` for my taste, but I'm just sticking with what I have…


backend/gke/force-clean-cluster.sh
```diff
@@ -49,3 +56,27 @@ for disk in $(jq -r ".disks[] | select(${DISK_SELECTOR:-true}) | .name" "${RESOU
...
+if [[ "${#recordsets[@]}" -gt 0 ]]; then
+    ${PREFIX} gcloud dns record-sets transaction start --zone="${dns_zone}"
+    for recordset in "${recordsets[@]}"; do
+        record_name="$(grep --perl-regexp --only-matching -- "--name='?\K[^']+(?=')" <<<"${recordset}")"
+        record_type="$(grep --perl-regexp --only-matching -- "--type='?\K[^']+(?=')" <<<"${recordset}")"
+        info "Removing DNS record set ${record_name} (type ${record_type})"
+        eval "${PREFIX} gcloud dns record-sets transaction remove ${recordset}"
+    done
+    info "Committing DNS changes…"
+    ${PREFIX} gcloud dns record-sets transaction execute --zone="${dns_zone}"
```
viccuad, line 81: I wonder if this is atomic, or if we would get into trouble when several jobs try to execute a transaction.

viccuad (approved):
Approving, leaving the merge to you. Seems like a good chunk of jq, but what can we do. I'm not super happy with `GKE_CLUSTER_NAME` assumption, see the other PRs.

mook-as:
Going to hit merge now, since at worst, it fails to find anything to delete.
Not happy with the reliance on cluster name either; see the other PR for comments (but basically, we don't have any method of passing other out-of-band information around).

PR #326, opened 2020-12-04
Don't remove KubeCF's ns manually, let quarks-operator do it

viccuad:
quarks-operator deletes the ns that is watching on its own, don't remove it manually.

Closes https://github.com/SUSE/catapult/issues/324


viccuad:
Tested and working, deployed kubecf on tf1, called `kubecf-clean` without this fix and it failed (Calling it again will succeed nevertheless).

Deployed kubecf again, and called `kubecf-clean` with this change, and it succeeds.

PR #329, opened 2021-05-28
Add belts & suspenders to wait_ns() function

jandubois:
Don't succeed if `kubectl` fails without any output. And for debugging purposes write the final output to stdout, just in case this is still not entirely correct.


mook-as (approved):
LGTM, though of course I haven't tested it… hopefully having extra output isn't going to break anything.

PR #330, opened 2021-05-28
wait_for_scf: make sure api is running, and ig jobs are gone

jandubois:
There can be a brief moment when the "ig" job has terminated, but none of the stateful sets have created a pod yet, where everything within the namespace seems to be ready. That's why we also have to wait for the "api" pod to exist.

There is a different scenario when one or more "ig" jobs have been launched by quarks, and they seem to be in "completed" status and everything appears to be "ready". But then one of the "ig" jobs restarts and terminates all the scf pods again. So we need to wait until all the "ig" jobs are gone.

### DIFF
--- a/include/func.sh
+++ b/include/func.sh
@@ -218,8 +218,8 @@ function wait_container_attached {
 # See wait_for_kubecf() below for a usage example, and why all this is needed.
 function wait_ns {
     local NAMESPACE=$1
-    local MUSTHAVE=$2
-    local MUSTNOT=$3
+    local MUSTHAVE=${2:-}
+    local MUSTNOT=${3:-}
 
     while true; do
         local output
--- a/include/func.sh
+++ b/include/func.sh
@@ -211,6 +211,11 @@ function wait_container_attached {
     done
 }
 
+# wait_ns NAMESPACE MUSTHAVE MUSTNOT
+# wait_ns waits for all pods in the NAMESPACE to be running or completed (and all containers to be running).
+# It also waits until an instance of the MUSTHAVE pod exist (if non-blank), and that no instances of MUSTNOT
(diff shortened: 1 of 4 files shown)


include/func.sh
```diff
@@ -212,15 +212,35 @@ function wait_container_attached {
 }
 
 function wait_ns {
```
mook-as, line 214: Might be a good idea to have some comments on how to use this thing. (Currently it only supports one must have and one must not, and must have must be provided if must not is desired.)

include/func.sh
```diff
@@ -377,10 +397,22 @@ wait_for_cf-operator() {
...
-    wait_ns scf
+    wait_for_scf
     #wait_for "kubectl delete -f ../kube/cf-operator/boshdeployment.yaml --namespace=scf"
     wait_for "kubectl delete -f ../kube/cf-operator/password.yaml --namespace=scf"
     if [[ "${DOCKER_REGISTRY}" == "registry.suse.com" ]]; then
       wait_for "kubectl delete -f ../kube/cf-operator/qstatefulset_tolerations.yaml --namespace=scf"
     fi
 }
+
+wait_for_scf() {
```
mook-as, line 408: If I remember correctly, `scf` doesn't have operator, just `kubecf` right? And I see that we didn't update `modules/scf/install.sh` to use the new code (probably because it doesn't have `ig` jobs).  Shouldn't this method be called `wait_for_kubecf` instead, then?

mook-as (approved):

include/func.sh
jandubois: Yes, I guess. It is `wait_ns scf` but actually waits for `kubecf`, which is totally not confusing or anything...

mook-as (approved):

